PR #23, opened 2015-09-14
glmer & horseshoe for bernoulli & binomial

jgabry:
@bgoodri This adds the glmer and horseshoe code to `bernoulli.stan` and `binomial.stan` and makes some related edits to `stan_glm.fit`. 

I’m doing this as a pull request to be safe. If it looks right to you then we can merge it into master.


jgabry:
@bgoodri also, if you already have this stuff implemented locally in a way you like we can just close this

bgoodri:
I'm working on this now. For the Bernoulli / Binomial case, the dispersion parameter should be fixed to 1 rather than estimated.

jgabry:
Yeah you're right. Glad you caught that.

PR #26, opened 2015-09-24
Fix posterior predict

jgabry:
Enables posterior predict for stan_glmer


jgabry:
@bgoodri Can you try this out and see if it breaks? I've tried a few examples and it seems to work. If it looks good to you we can merge.

PR #53, opened 2016-01-14
Make HTML vignettes self-contained so they can work on CRAN

yihui:
I'm not sure why you preferred `self_contained = FALSE`, but for vignettes to be viewed correctly on CRAN. It will be much easier for CRAN maintainers to copy your vignettes without taking care of external image files.

Note I also changed the R CMD INSTALL option `--precleanOOPS` to `--preclean`. I don't know what the former means. It does not seem to be a standard option.

The default file size of the `html_document` is large due to the Bootstrap dependencies, and you can reduce the vignette sizes significantly by using `html_vignette`. Let me know if you want to do that and I can amend this PR.


bgoodri:
Thanks @yihui ! The motivation for self_contained = FALSE was that CRAN doesn't like more than 5MB of vignettes but we wanted to put a lot of images based on MCMC output into the HTML vignettes. With svglite and external images, this got us down to 3.5 MB but everything else we tried was putting us over. I can't remember if we tried `html_vignette`, but most of the bulk was due to the images. I had hoped that the script CRAN uses to push the vignettes to the website would read vignettes/.install_extras but apparently it does not.

The `--precleanOOPS` thing was just a hack we came up with accelerate development because RStudio wanted to add the `--preclean` argument unless it thought the option was already (mis)specified. With the `--preclean` argument, it rebuilds the autogenerated content in rstanarm/src/ even when the developer does not need to at that moment. R CMD INSTALL just ignores the `--precleanOOPS` argument with a warning.

yihui:
Thanks for explaining the `--precleanOOPS` hack. That is interesting :)

For the vignette size problem, I have added a few more commits to help. Basically,

1. When `self_contained = TRUE`, there is no need to ship the SVG images with the source package, since they will be generated at the build time; putting them in the source package will only increase the size of the `vignettes/` directory.
1. `html_vignette()` removed the Bootstrap dependencies, hence gives smaller vignette output than `html_document()`;
1. The default png device generates smaller images (in file size) than svg, so I removed the svg device.

Now the size of vignette output (`inst/doc`) is about 1.2Mb, and I think that should make CRAN happy.

jgabry:
Thanks a lot Yihui! This is really helpful.

yihui:
No problem. Let me know if you need anything else. This PR should be complete and ready to merge now.

bgoodri:
Merged now, thanks!

PR #59, opened 2016-01-18
Bugfix issue 49

### LINKED ISSUE #49: Wrong fixed effects estimates when running stan_glm without an intercept
I get weird fixed effects estimates when fitting a model without an intercept:

```R
library(rstanarm)
devtools::install_github("paul-buerkner/brms")
library(brms)
foo1 <- stan_glm(mpg ~ -1 + ., data=mtcars, chains=2, iter=2000, warmup=500)
foo1

foo2 <- brm(mpg ~ -1 + ., data=mtcars, chains=2, iter=2000, warmup=500)
foo2

foo3 <- glm(mpg ~ -1 + ., data = mtcars)
summary(foo3)
```
The results of `foo2` and `foo3` closely match each other while `foo1` produces very different results. Especially the  `MD_SD` (and `SD`) estimates are a bit off. This may be a result of centering the fixed effects design matrix X regardless of whether an intercept is present or not. I believe that centering is invalid in models without an intercept (at least when not corrected otherwise) but I may be mistaken.

jgabry:
@bgoodri I added a test for stan_glm for a model without the intercept. If the travis checks pass (everything already passes locally for me) I think this is read to merge. 

Fixes #49

### DIFF
--- a/exec/binomial.stan
+++ b/exec/binomial.stan
@@ -143,6 +143,8 @@ generated quantities {
         alpha[1] <- alpha[1] - shift;
       }
     }
+    #include "eta_no_intercept.txt"
+    
     pi <- linkinv_binom(eta, link);
     for (n in 1:N) mean_PPD <- mean_PPD + binomial_rng(trials[n], pi[n]);
     mean_PPD <- mean_PPD / N;
--- a/exec/binomial.stan
+++ b/exec/binomial.stan
@@ -110,6 +110,7 @@ model {
     if (link != 4) eta <- eta + gamma[1];
     else eta <- gamma[1] + eta - max(eta);
   }
+  #include "eta_no_intercept.txt"
   
   // Log-likelihood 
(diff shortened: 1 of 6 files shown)


bgoodri:
Also remember to put an entry into NEWS.

jgabry:
Looks like these tests are failing, although I'm not sure why. I think the failing tests are unrelated to these commits.

bgoodri:
Yeah, I merged it anyway. In the R code, we should also prohibit models
without intercepts when the link function is insufficient to ensure that
the linear predictors are admissible. In a few places in the Stan code, we
are relying on the (existence of the) intercept to be sufficiently extreme
to impose an inequality constraint on the linear predictors.

jgabry:
So like  link="inverse" for Gamma and link="identity" for a bunch of them? Maybe a few others too.

bgoodri:
Yeah. And link="log" for binomial / bernoulli.

PR #62, opened 2016-01-21
Allow t-distributed errors in stan_glm, stan_glmer

### LINKED ISSUE #35: allow using student_t instead of gaussian for likelihood
For `stan_glm` with a suitable outcome (and maybe `stan_lmer`), we should provide an option to use the t distribution as the likelihood instead of a Gaussian. 

Perhaps we just follow BDA3 17.2 and use the fact that 

y_i ~ t_v(m, s^2) 

is equivalent to 

y_i | V_i ~ N(m, V_i)
V_i ~ scaled-inv-chi-sq(v, s^2)

jgabry:
resolves #35


jgabry:
A few issues before we merge this:

* Prior on the degrees of freedom parameter

Right now I'm just using [Aki's recommendation of gamma(2, 0.1) as a default](https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations), and I've hardcoded it. Should we instead provide a mechanism through which the user can specify the prior?

* Name for the family object

What should we call the new family object? `student_t` is taken by the function used for specifying student t priors, so right now the new family object/function is called `t_family`, but that's quite different from the other family objects, whose names don't contain the word "family".

* Testing

How should we test these models? There's a `tlm` function in the **hett** package that we could test against (if we null out the priors and use optimization), but that would require adding **hett** to the list of packages in Suggests in order to use it to test this one feature. Is there anything in any of the packages we're already using? The `rlm` function in **MASS** does robust regression but doesn't seem to do it using t-distributed errors.

bgoodri:
I think it is fine to start by only supporting a Gamma prior for the degrees of freedom, but there should be a mechanism for the user to choose the hyperparameters. This is what I did in `stan_polr()` in the scobit case. For testing, we could just write the likelihood as an R function, call `optim()`, and compare it to `optimizing()`.

jgabry:
Ok I'll do all of those things. Can you review the rest of the pull request
and check my work in the Stan code and on loo and posterior_predict? It
will probably take you 5 min and I think an extra pair of eyes on this
would be good before we include it.

bgoodri:
One meta-comment: I don't like referring to this model as "robust regression" even though that phrase is used in ARM and in the Stan manual. The word robust is used in lots of different ways. For users of rstanarm, I could easily see people confusing what is done here with robust standard errors, MASS::rlm, robust::lmRob, or other things. I think it would just be better to say it is a linear models where the errors are believed to be distributed Student t conditional on the predictors, which is equivalent to believing the errors are normal but have different standard deviations that themselves have the scaled inverse chi-squared distribution.

.travis.yml
```diff
@@ -19,12 +19,12 @@ r_binary_packages:
   - Rcpp
   - coda
   - BH
-  - RcppEigen
   - RInside
   - ggplot2
   - gridExtra
   - rstan
   - shinystan
+#  - RcppEigen
```
bgoodri: Can switch back to binary now

R/families.R
```diff
@@ -50,3 +51,44 @@ neg_binomial_2 <- function(link = "log") {
...
+#' 
+#' @name t_family
+#' @export
+#' @templateVar armRef (Ch. 6)
+#' @template reference-gelman-hill
+#' @param link The same as for \code{\link[stats]{gaussian}}.
+#' @return An object of class \code{\link[stats]{family}} very similar to
+#'   that of \code{\link[stats]{gaussian}} but with a different family name.
+#' @examples 
+#' \dontrun{
```
bgoodri: Would rather run a five second example

R/loo.R
```diff
@@ -202,6 +206,11 @@ ll_args <- function(object) {
   val <- dnorm(data$y, mean = .mu(data,draws), sd = draws$sigma, log = TRUE)
   .weighted(val, data$weights)
 }
+.ll_t_family_i <- function(i, data, draws) {
+  x <- (data$y - .mu(data, draws)) / draws$scale
+  val <- dt(x, df = draws$df, log = TRUE) - log(abs(draws$scale))
```
bgoodri: The `abs()` is unnecessary

jgabry:
Ok, I think I addressed all of your suggestions in these most recent commits. But I'm not sure what the best way is to allow the user to set the hyperparameters for the gamma prior on the df. Right now I decided to add it to what can be specified using the prior_options() function, but I'm not sure that's the best way to go. Any preference?

jgabry:
Although now a bunch of tests are failing, so maybe something got messed up. I'll see if I can sort it out.

bgoodri:
It feels like the shape and rate hyperparameters should be optional
arguments in the call to t_family(), although we haven't been doing things
like that with the other families. But I don't especially like how
prior_options() is becoming a bunch of miscellaneous stuff.

jgabry:
Yeah I thought about adding them to t_family but it doesn't fit with the
other families. I don't like prior_options either, but I don't have a
better suggestion unless we want to change all of the family objects to
allow setting hyperparameter values. That would be nice but then we
couldn't use the standard family names.

jgabry:
Also, the tests are passing now but the travis build is failing because it's missing all of the documentation for some reason. I have no idea why this is a problem all of a sudden. Here's the [relevant part from the travis log](https://travis-ci.org/stan-dev/rstanarm/builds/104372382):

<pre> 
* checking for missing documentation entries ... WARNING
Undocumented code objects:
  ‘R2’ ‘VarCorr’ ‘cauchy’ ‘compare’ ‘decov’ ‘dirichlet’ ‘example_model’
  ‘fixef’ ‘get_x’ ‘get_y’ ‘get_z’ ‘hs’ ‘hs_plus’ ‘launch_shinystan’
  ‘log_lik’ ‘loo’ ‘neg_binomial_2’ ‘ngrps’ ‘normal’
  ‘posterior_interval’ ‘posterior_predict’ ‘pp_check’ ‘ppcheck’
  ‘prior_options’ ‘ranef’ ‘se’ ‘sigma’ ‘stan_aov’ ‘stan_gamm4’
  ‘stan_glm’ ‘stan_glm.fit’ ‘stan_glm.nb’ ‘stan_glmer’ ‘stan_glmer.nb’
  ‘stan_lm’ ‘stan_lm.fit’ ‘stan_lm.wfit’ ‘stan_lmer’ ‘stan_polr’
  ‘stan_polr.fit’ ‘student_t’ ‘t_family’ ‘waic’
Undocumented data sets:
  ‘example_model’
All user-level objects in a package should have documentation entries.
See chapter ‘Writing R documentation files’ in the ‘Writing R
Extensions’ manual.
</pre>

bgoodri:
Again, I'm not sure there is much harm to masking stats::gaussian since any
calls of the form gaussian(link = "something") would continue to work. It
is just that the resulting list would contain unused default values of the
hyperparameters in the prior distribution for sigma.

bgoodri:
Or we can make a rstanarm_family(name, link, ...) function and use it
everywhere that a user can specify if they want non-default hyperparameters.

jgabry:
I like that option better than masking all the stats family functions.

bgoodri:
Ok, let's do it. I wish we had thought of this before I did scobit, but
whatever. We should also have a description of the gamma distribution in
priors.R even if there is no gamma_prior() function.

jgabry:
Can we do the scobit this way too and deprecate the shape and rate
parameters to stan_polr?

bgoodri:
Yeah

jgabry:
Ok, I like this idea. I think doing everything this way will allow for much
more flexibility going forward. The prior_options thing did always feel
ugly.

betanalpha:
Let me advise severe caution in these models.  Student-t measurement
models are likely to induce heavy-tailed posteriors which HMC can have
trouble exploring.  More specifically, it can get stuck out in the tails never
to return to the bulk around the mode (less so that Random Walk Metropolis
but still enough to ruin the fit).  Before releasing these to users I recommend
_extensive_ testing with many chains to ensure that you can at least
recover simulated values.

jgabry:
Thanks Michael. I'm concerned about this but so far the only examples I've
found where the parameters are not recovered lead to our pretty
severe convergence warnings ("[...] Do not analyze results!") being
triggered. That said, it wouldn't necessarily be clear to the user that
this was the particular problem they were facing, and I've only had this
implemented for a few days so there are plenty of more models I should
test. If it does end up being robust enough to add I will add make sure to
highlight this concern in the documentation.

If we do end up seeing too many of these cases then I have no problem
continuing to not allow t-distributed errors or plastering the screen
with threats of death and despair.

Also the Pareto k parameter estimates obtained from running loo on these
models is also a useful diagnostic in these cases.

betanalpha:
Actually the pathologies due to heavy-tailed posteriors shouldn't manifest as divergences.  It could be that all of your examples have enough data to induce reasonable posteriors, but just keep an eye out and be sure to use plenty of chains.

jgabry:
I was actually referring to warnings that are triggered by rhat values, but
that's interesting to know about the divergences. I hadn't thought about it.

But it should tend to manifest as poor mixing right? That's why I've been
encouraged by the rhats so far (in tandem with the parameter recovery).

betanalpha:
Yeah, good Rhats with multiple chains and recovering the simulated
values are solid evidence that everything is fitting okay.  Although it
would be great to see really disperse initial conditions, just in case.

jgabry:
@bgoodri I think this is ready to merge. Can you give it a try and take a look at the doc and then we can merge if it looks good? Actually, did you also want to use `rstanarm_family()` for `stan_polr` also? 

----
### Contents: 

* `t_family` for GLMs with t-distributed errors (tested with `stan_glm` against `stats::optim`)
* `rstanarm_family` for specifying family, link, and family-specific hyperparameters

bgoodri:
Aside from resolving the merge conflicts, is there anything more that we wanted to do here? We should be checking the energy diagnostic, not just in models with a Student t likelihood but maybe it is especially applicable in that case.

bgoodri:
Now the BFMI thing is checked from rstan 2.13.x

jgabry:
I think we should give a bit more thought to how to do the
rstanarm_family() thing. I think what's in the PR is probably similar to
what we want but it's been a while since we've thought about it and it's a
pretty big change (for users at least).. So we should make sure it's really
how we want it to be going forward.

I haven't looked at it for a while so I should go through the PR again.
I'll do that soon and I can fix conflicts when I do that.

avehtari:
- real<lower=0> nu[family == 4];
  no sense to have nu lower than 1, which corresponds to Cauchy
  even that is sometimes problematic and lower=2 is also sensible
- Is it ok to extend glm to refer models using other than exponential family distributions as observation models? Or should there be another function name for these?
- Even if stan_glm would be extended, family refers to exponential family and it seems strange to have family=student_t. Maybe something else?
- If family is used, could it be mentioned somewhere that in Bayesian modeling these correspond to observation model distrbutions?
- I hate family="t_family".
- As comparison surv package has
  survreg(formula, data, weights, subset, na.action, dist="weibull", ...)
where dist can be
  "extreme" "logistic" "gaussian" "rayleigh" "loggaussian" "lognormal"
  "weibull" "exponential" "loglogistic" "t"

- Would it be better to leave stan_glm to resemble glm, and then make
  stan_reg (or something similar) with dist argument and allow wide
  variety of distributions (not families)?

- and hopefully at some point there would be also stan_survreg function

bgoodri:
On Mon, Nov 28, 2016 at 3:51 PM, Aki Vehtari <notifications@github.com>
wrote:

>
>    -
>
>    real<lower=0> nu[family == 4];
>    no sense to have nu lower than 1, which corresponds to Cauchy
>    even that is sometimes problematic and lower=2 is also sensible
>
> So, truncated prior on nu if it is estimated?



>
>    - Is it ok to extend glm to refer models using other than exponential
>    family distributions as observation models? Or should there be another
>    function name for these?
>    -
>
>    Even if stan_glm would be extended, family refers to exponential
>    family and it seems strange to have family=student_t. Maybe something else?
>    -
>
>    If family is used, could it be mentioned somewhere that in Bayesian
>    modeling these correspond to observation model distrbutions?
>    -
>
>    I hate family="t_family".
>
> Today I am leaning more toward having a stan_treg() function with no
family argument.

>
>    -
>
>    As comparison surv package has
>    survreg(formula, data, weights, subset, na.action, dist="weibull", ...)
>    where dist can be
>    "extreme" "logistic" "gaussian" "rayleigh" "loggaussian" "lognormal"
>    "weibull" "exponential" "loglogistic" "t"
>    -
>
>    Would it be better to leave stan_glm to resemble glm, and then make
>    stan_reg (or something similar) with dist argument and allow wide
>    variety of distributions (not families)?
>    -
>
>    and hopefully at some point there would be also stan_survreg function
>
> We are in the process of putting a joint survival & tumor size model into
rstanarm that takes a Surv() ~ ... formula, so that will get us some of the
survreg infrastructure.

avehtari:
On 28.11.2016 23:08, bgoodri wrote:
> So, truncated prior on nu if it is estimated?

Yes.

> Today I am leaning more toward having a stan_treg() function with no
> family argument.

Why not stan_reg with dist argument? It would be then more natural way to add 
all sensible regression distributions already in stan math.

> We are in the process of putting a joint survival & tumor size model into
> rstanarm that takes a Surv() ~ ... formula, so that will get us some of the
> survreg infrastructure.

Great!

Aki

bgoodri:
> Why not stan_reg with dist argument? It would be then more natural way to add all sensible regression distributions already in stan math.

I wouldn't want users to be put in a position of deciding whether to use `stan_glm` or `stan_reg` to do a Gaussian, Poisson, etc. I guess the issue is how many additional regression models do we anticipate adding to `stan_reg`? For things like survival models, I imagine there would be a `stan_survreg`. But there doesn't seem to be a widely-used package / function that implements regression with a Student t likelihood.

PR #65, opened 2016-01-31
move .txt files in exec/ to inst/chunks

bgoodri:
and change their extension to .stan so that GitHub recognizes them as Stan code


jgabry:
Any idea why this caused the build to start failing again apparently?

bgoodri:
No. It is still installing roxygen2, so I don't know why the documentation
is missing.

jgabry:
If I do `R CMD INSTALL --preclean rstanarm/` I now get the following error 

```{r}
Error in file(con, "r") : cannot open the connection
Calls: <Anonymous> ... lapply -> FUN -> <Anonymous> -> append -> readLines -> file
In addition: Warning message:
In file(con, "r") :
  cannot open file 'inst/chunks/make_eta_bern.stan': No such file or directory
```

although it seems to continue after that.

bgoodri:
I fixed that now

jgabry:
Ok looks like that fixed it. I don't get the error anymore.

bgoodri:
Do you have any idea why Travis says "R binary packages not supported on osx, falling back to source install" at https://travis-ci.org/stan-dev/rstanarm/builds/106085935#L143 . Looking at the contents of the tarball, the man/ directory is, in fact, missing.

jgabry:
I didn't know that about binary packages, but if you look at https://docs.travis-ci.com/user/languages/r#Dependencies under `r_binary_packages` it does say

> On OSX builds, these packages are installed from source.

bgoodri:
Travis seems to have fixed itself without any actual changes to the R code.

jgabry:
Yeah it's rather mysterious

PR #89, opened 2016-04-08
Feature/plot prior vs posterior

jgabry:


jgabry:
@bgoodri Are we ready to merge this?

bgoodri:
Yeah, it just needs a test I think.

jgabry:
There are already a few tests for it at the end of the [`test_plots.R`](https://github.com/stan-dev/rstanarm/blob/feature/plot-prior-vs-posterior/tests/testthat/test_plots.R) file, although they're basically just checking for errors and that a ggplot object is returned, not really the content of the plot.

PR #93, opened 2016-04-23
purge example_model

bgoodri:
To use the example_model in examples, you have to first call
if (!exists("example_model")) example(example_model)


jgabry:
Looks fine, but what prompted the change?

bgoodri:
There seem to be more than more issues where the rstanarm namespace is not
found, which I am hoping is due to the old example_model needing the
rstanarm namespace in order to be put into the rstanarm namespace. Also, it
is now easier to keep in sync as rstan changes. For example, pairs from
2.10.0 is flaky if example_model was created with 2.9.0-3 because of the
renaming of n_divergent__ to divergent__ etc.

jgabry:
On Saturday, April 23, 2016, bgoodri <notifications@github.com> wrote:

> There seem to be more than more issues where the rstanarm namespace is not
> found, which I am hoping is due to the old example_model needing the
> rstanarm namespace in order to be put into the rstanarm namespace. Also, it
> is now easier to keep in sync as rstan changes. For example, pairs from
> 2.10.0 is flaky if example_model was created with 2.9.0-3 because of the
> renaming of n_divergent__ to divergent__ etc.


Ok makes sense. Also, if that n_divergent change is going to be in the next
rstan then I need to also remember to update shinystan or it might (will
probably) break. But then which shinystan version you need will depend on
which rstan you have. That's annoying but will be unavoidable I guess.
>
>

bgoodri:
It would be easy enough to have the next shinystan work with either 2.9.0-3
or 2.10.0 but hard for the current shinystan to work with 2.10.0.

jgabry:
Yeah, I can have it check the rstan version but if you update rstan you'll
have to update shinystan. Maybe we should specify a version of shinystan in
Suggests for rstan (if we don't already). If someone updates rstan without
updating shinystan then they will probably just see a mysterious error from
shinystan and we'll probably get questions and/or bug reports.

PR #99, opened 2016-06-10
Feature/sparse x

bgoodri:
@jgabry Look at this. It is a bit hard to unit test, although I changed some tests involving `stan_glm.fit` to `sparse = TRUE` to verify that it does not throw errors anymore. The main thing is that when the user specifies `sparse = TRUE`, then we can't center the `X` matrix without destroying all the exact zeros, and the intercept is a regular intercept. So, if there is an informative prior on the intercept, then the results are not exactly the same (for the same seed) depending on whether `sparse = TRUE` or `sparse = FALSE`. Also, the performance seems really bad even for a high degree of sparseness, so this is really only about making it possible to use rstanarm for datasets that are too large when represented in the dense format.


jgabry:
Cool. Will go through it this weekend.

PR #108, opened 2016-07-16
Feature/lm list

bgoodri:
@jgabry Is there anything else to implement for this partially pooled linear model before merging? It is failing on travis for mysterious reasons but passing locally.

Is the stuff you were doing to do `pp_check` by group on a branch somewhere? We could use something like that for this. `example("lmList", package = "lme4")` has a nice (lattice-based) plot.


bgoodri:
Upon further review, the problem is that we have a massive memory leak somewhere.

jgabry:
Haven't had a chance to look at this yet. Probably Thursday or Friday at
this point.

bgoodri:
OK, it is fundamentally passing on Travis now

jgabry:
@bgoodri Just started going through this. One thing I noticed is that it doesn't seem to work with `loo` when `k_threshold` is specified: 

```{r}
Error: grouping factors must have > 1 sampled level
Called from: checkNlevels(reTrms$flist, n = n, control)
```
which comes from these new lines in `.pp_data` (which gets called by `log_lik` when `newdata` is specified):
```
  if (inherits(object, "lmList")) {
    f <- as.character(object$formula)
    f <- as.formula(paste(f[2], "~", "-1 + (", f[3], ")"))
    g <- glFormula(f, newdata) # error happens here
```

bgoodri:
Looks like I need to pass a glmerControl() to glFormula() like we do in
stan_glmer()

jgabry:
Also, do you know why this gives such crazy fitted values (~ -500 mpg) for some obs?

```
> post <- stan_lmList(mpg ~ disp + wt + hp | cyl, data = mtcars, 
                    prior = R2(0.25),cores = 4)

> post$fitted.values
 [1] -445.53464662 -446.08058883   34.64203080 -627.35905689    4.99964496 -559.32449712
 [7]    4.32078553   32.79270824   33.07806188 -481.47190037 -481.47190037    8.84858062
[13]    8.85052841    8.85024197   -0.49029885   -0.03220982    0.74953141   36.04257352
[19]   36.17966384   36.40331209   34.06388449    7.18722826    7.83633447    4.78253823
[25]    3.14412690   36.02574804   34.05075351   35.25456873    4.55598688 -519.11948085
[31]    6.18239069   34.02475271
```

bgoodri:
Those are the 6 cylinder cars but I don't know what it isn't matching up
more closely with colMeans(posterior_predict())

jgabry:
Yeah, two issues I think. One is that it's so different from `posterior_predict`. The other is that the fitted value is __negative__ 500 miles per gallon, which seems impossible since I would think 0 is a logical lower bound on mpg. So maybe negative 5 could make sense for a noisy estimate but negative 500?

bgoodri:
I think the ordering in $coefficients is wrong.

PR #113, opened 2016-08-24
Bugfix/factors with spaces

jgabry:
* Replaces spaces with underscores in factor levels in `.pp_data_mer_Z` to match what `pad_reTrms` does for `stan_glm.fit`

* Adds `XZ` argument to `posterior_linpred`. If `XZ=TRUE` then `X` (or `cbind(X,Z)` if applicable) is returned instead of the posterior of the linear predictor 

@bgoodri Can you take a peek at this and see if it looks ok to you? There’s only a few lines of code to look at.


jgabry:
I didn't know what to call the argument for returning X:Z so I just called it `XZ`. If you have a better name for it we can definitely change it

PR #116, opened 2016-09-14
fix for gaussian(link="log")

jgabry:
In `continuous.stan`, for gaussian family with log link use exponentiated linear predictor with normal distribution rather than the lognormal distribution. 

@bgoodri This look right to you?


bgoodri:
Yes. Is all the R code already doing the right thing?

jgabry:
I think so

bgoodri:
OK. Merge it. Travis may fail for unrelated reasons.

jgabry:
Ok I'll double check the R code and then merge it

PR #123, opened 2016-10-13
predictive intervals and errors

jgabry:
@bgoodri This PR introduces two new convenience functions. They are both essentially wrappers for `posterior_predict`. 
- `predictive_error` computes in-sample (y - yrep) or out-of-sample (y - ytilde) predictive errors
- `predictive_interval` is essentially the same as `posterior_interval`, but instead of for model parameters it returns intervals for predictions

I wasn’t sure what to do for either of these functions with ordinal models (for which posterior_predict returns a character matrix and these functions probably don’t make sense), so they currently throw an error for models fit using `stan_polr`.


jgabry:
Alternative, instead of having `predictive_interval` and `predictive_error` call `posterior_predict` internally, they could take the object returned by `posterior_predict` as their input. This would avoid `posterior_predict` needing to be called multiple times but it means that these functions would no longer take `stanreg` objects as input, and I guess it's kind of nice that they take them as inputs and it would only be inconvenient for really huge datasets having to call `posterior_predict` multiple times. @bgoodri do you have a preference?

bgoodri:
If it is a S3 method, then I don't see why you couldn't have both.

jgabry:
True, we could definitely have both if we do want both.

jgabry:
closing as it will help avoid conflicts to include this in the PR for using __rstantools__

PR #125, opened 2016-10-27
Introduce prior_summary

jgabry:
This PR adds a prior_summary function for extracting and/or printing a summary of the priors specified by the user (and those actually used by rstanarm if there are internally adjusted scales). 

To do this, I also added internal functions in stan_glm.fit, stan_biglm.fit, and stan_polr.fit that keep track of this information, which is eventually stored in the ‘prior.info’ slot of the resulting stanreg object. The prior_summary function returns the prior.info slot and its print method formats it nicely for the screen.


jgabry:
closing as it will help avoid conflicts to include this in the PR for using __rstantools__

PR #126, opened 2016-10-31
Use rstantools pkg

jgabry:
This PR: 

* imports some S3 generics from rstantools instead of having them in rstanarm
* for some generics in rstantools there are not currently methods in rstanarm and this PR adds those. these include `prior_summary`, `predictive_interval`, and `predictive_error`. For the last two, methods that accept output from `posterior_predict` are also included along with methods for stanreg objects.


bgoodri:
Is this ready to merge? We should try to get a rstanarm release onto CRAN before the Nov 22 webinar.

jgabry:
I think it's ready. Do the S3 methods work properly for you when the generics are imported from __rstantools__? Also are you ok with how `prior_summary` displays the information?

Also, I suppose we/I should release __rstantools__ at least a few days before we release __rstanarm__, right?

bgoodri:
I will check tonight. Also, we need to do the same with bayesplot right?

jgabry:
Yeah same with bayesplot. I started working on that already but haven't finished yet.

bgoodri:
This is passing all tests, so it is probably okay to upload rstantools to CRAN next week and then merge this PR. It sounds like doing the same with bayesplot may be more of a stretch.

PR #130, opened 2016-11-20
Feature/betareg

imadmali:


R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+                              algorithm = c("sampling", "optimizing", "meanfield", "fullrank"),
+                              adapt_delta = NULL, QR = FALSE, sparse = FALSE) {
+  
+  # lots of tedious but simple stuff including standata which is a big list to pass to data {}
+  # process the prior information like stan_glm.fit() does
+  
+  algorithm <- match.arg(algorithm)
+
+  # determine whether the user has passed a matrix for the percision model (z)
+  if (length(link.phi) == 1) {
```
jgabry: Additionally, what happens if a user omits the `z` argument when calling `stan_betareg.fit` directly? Will it work? Without trying it my hunch is that it will error somewhere. That means that the user needs to specify something for `z` even if they don't have it in their model?

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+
+  if (nvars_z == 0) {
+      prior_mean_z <- double()
+      prior_scale_z <- double()
+      prior_df_z <- integer()
+  }
+
+  # create entries in the data block of the .stan file
+  standata <- nlist(
+    N = nrow(xtemp), K = ncol(xtemp), xbar = as.array(xbar), dense_X = !sparse, # TRUE, sparse = FALSE,
```
jgabry: I don't understand the comment `# TRUE, sparse = FALSE,`

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+    v_X = integer(), 
+    u_X = integer(), 
+    y = y, 
+    prior_PD, has_intercept, family = 4L, link = link_num, 
+    prior_dist, prior_mean, prior_scale = as.array(pmin(.Machine$double.xmax, prior_scale)), prior_df,
+    prior_dist_for_intercept, prior_mean_for_intercept = c(prior_mean_for_intercept), 
+    prior_scale_for_intercept = min(.Machine$double.xmax, prior_scale_for_intercept), 
+    prior_df_for_intercept = c(prior_df_for_intercept),
+    prior_scale_for_dispersion = prior_scale_for_dispersion %ORifINF% 0,
+    has_weights = length(weights) > 0, weights = double(),
```
jgabry: why `weights = double()` instead of whatever weights the user specified?

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+    has_offset = length(offset) > 0, offset = double(),
```
jgabry: same question about `offset` as with `weights` above

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+    l = integer(), 
+    q = 0L, 
+    len_theta_L = 0L, shape = double(), scale = double(), 
+    len_concentration = 0L, concentration = double(),
+    len_regularization = 0L, regularization = double(),
+    num_non_zero = 0L, 
+    w = double(), 
+    v = integer(), 
+    u = integer(),
+    z_dim = nvars_z, # ncol(z),
```
jgabry: is this comment indicating that `nvars_z` is equal to `ncol(z)`? if not then it's confusing

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+    pars <- c(if (has_intercept) "alpha", "beta", "omega_int", "omega", "mean_PPD")
+  }
+  else {
+    pars <- c(if (has_intercept) "alpha", "beta", "dispersion", "mean_PPD")
+  }
+
+  if (algorithm == "optimizing") {
+    out <- optimizing(stanfit, data = standata, draws = 1000, constrained = TRUE, ...)
+    out$par <- out$par[!grepl("eta_z", names(out$par))] # might need fixing
+    out$theta_tilde <- out$theta_tilde[,!grepl("eta_z", colnames(out$theta_tilde))] # might need fixing
```
jgabry: "might need fixing" ??

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+    new_names <- names(out$par)
+    mark <- grepl("^beta\\[[[:digit:]]+\\]$", new_names)
+    new_names[mark] <- colnames(xtemp)
+    new_names[new_names == "alpha[1]"] <- "(Intercept)"
+    if (Z_true == 1) {
+      new_names[new_names == "omega_int[1]"] <- "(phi)_(Intercept)"
+      mark_z <- grepl("^omega\\[[[:digit:]]+\\]$", new_names) # "^omega\\[[[:digit:]]+\\]$"
```
jgabry: the comment looks identical to what's already inside `grepl`

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+        object = stanfit, 
+        prior = prior, 
+        user_dots = list(...), 
+        user_adapt_delta = adapt_delta, 
+        data = standata, 
+        pars = pars, 
+        show_messages = FALSE)
+      stanfit <- do.call(sampling, sampling_args)
+    }
+    else if (algorithm == "meanfield") { # FIXME
```
jgabry: what is the FIXME indicating needs to be fixed?

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+      stanfit <- rstan::vb(stanfit, pars = pars, data = standata,
+                           algorithm = algorithm, init = 0.001, ...)
+    }
+    else if (algorithm == "fullrank") { # FIXME
```
jgabry: same question regarding FIXME

R/stan_betareg.fit.R
```diff
@@ -0,0 +1,225 @@
...
+  if (length(link.phi) == 1) {
+    Z_true <- 1
+  }
+  else {
+    Z_true <- 0
+  }
+
+  # no family argument
+  famname <- "beta"
+  is_beta <- is.beta(famname)
```
jgabry: is this `is_beta` object used anywhere?

R/stanreg.R
```diff
@@ -31,6 +31,11 @@ stanreg <- function(object) {
   nvars <- ncol(x)
   nobs <- NROW(y)
   ynames <- if (is.matrix(y)) rownames(y) else names(y)
+  if (family$family == "beta") {
```
jgabry: I think you made an `is.beta` function, so could be `is.beta(family$family)`

R/stanreg.R
```diff
@@ -43,9 +48,20 @@ stanreg <- function(object) {
     ses <- apply(stanmat[, xnms, drop = FALSE], 2L, mad)
     rank <- qr(x, tol = .Machine$double.eps, LAPACK = TRUE)$rank
     df.residual <- nobs - sum(object$weights == 0) - rank
+    if (family$family == "beta") {
```
jgabry: `is.beta(family$family)`

R/stanreg.R
```diff
@@ -43,9 +48,20 @@ stanreg <- function(object) {
...
+      if (length(colnames(z)) == 1)
+        coefs_z <- apply(stanmat[, grepl("(phi)", colnames(stanmat), fixed = TRUE), drop = FALSE], 2L, median)
+      else
+        coefs_z <- apply(stanmat[, paste0("(phi)_",colnames(z)), drop = FALSE], 2L, median)
+    }
   } else {
     stan_summary <- make_stan_summary(stanfit)
     coefs <- stan_summary[1:nvars, select_median(object$algorithm)]
+    if (family$family == "beta") {
```
jgabry: `is.beta`

R/stanreg.R
```diff
@@ -73,6 +89,11 @@ stanreg <- function(object) {
     residuals <- ytmp - mu
   }
   names(eta) <- names(mu) <- names(residuals) <- ynames
+  if (family$family == "beta") {
```
jgabry: actually if it's gonna be checked this many times might as well just do `is_betareg <- is.beta(family$family)` once towards the top and then just check `is_betareg` when you need to.

R/stanreg.R
```diff
@@ -73,6 +89,11 @@ stanreg <- function(object) {
...
+    eta_z <- linear_predictor(coefs_z, z, object$offset)
+    phi <- family_phi$linkinv(eta_z)
+    # residuals <- ytmp - rbeta(length(eta), mu * phi, (1 - mu) * phi) # this is not what betareg does
```
jgabry: what does betareg do? for `stan_glm` we don't do what `glm` does. we always to residuals on the outcome scale (check out the Details section of `?"stanreg-methods"`).

R/stanreg.R
```diff
@@ -106,6 +127,13 @@ stanreg <- function(object) {
     out$asymptotic_sampling_dist <- stanmat
   if (mer) 
     out$glmod <- object$glmod
+  if (family$family == "beta") {
```
jgabry: same

exec/continuous.stan
```diff
@@ -62,9 +62,59 @@ functions {
...
+      for(n in 1:rows(eta)) mu[n] = cauchy_cdf(eta[n], 0.0, 1.0);
+    else if (link == 5)  // log 
+      for(n in 1:rows(eta)) mu[n] = exp(eta[n]);
+    else if (link == 6) // loglog
+      for(n in 1:rows(eta)) mu[n] = 1-inv_cloglog(-eta[n]); 
+      
+    for (n in 1:rows(mu)) { 
+      //FIXME: maybe check this in tests but not in released version?
+      if (mu[n] < 0 || mu[n] > 1)
+        reject("mu needs to be between 0 and 1")
```
jgabry: @bgoodri should we remove this loop over `N` checking for valid `mu` for beta regression?

exec/continuous.stan
```diff
@@ -324,6 +497,33 @@ generated quantities {
...
+        yrep = beta_rng(eta[n] * eta_z[n], (1 - eta[n]) * eta_z[n]);
+          if (is_nan(yrep) == 1) {
+            mean_PPD = mean_PPD;
+            nan_count = nan_count + 1;
+          }
+          else if (is_nan(yrep) == 0) {
+            mean_PPD = mean_PPD + yrep; 
+          }
+          else {
+            print("something went wrong");
```
jgabry: what's this print statement for?

tests/testthat/test_stan_functions.R
```diff
@@ -411,3 +411,26 @@ test_that("the Stan equivalent of lme4's Z %*% b works", {
   test_lme4(glFormula(angle ~ recipe + temp + (1|recipe:replicate), data = cake)$reTrms)
 })
 
+# beta
+links <- c("logit", "probit", "cloglog", "cauchit", "log")
+
+context("Beta")
+test_that("linkinv_bern returns expected results", {
```
jgabry: i think this should be `linkinv_beta` not `linkinv_bern`

vignettes/betareg.Rmd
```diff
@@ -0,0 +1,195 @@
...
+```{r simulated-data, fig.height=5}
+dat <- list()
+dat$eta <- c(1, -0.2)
+dat$gamma <- c(1.8, 0.4)
+dat$N <- 500
+dat$x <- rnorm(dat$N, 2, 2)
+dat$z <- rnorm(dat$N, 2, 2)
+dat$mu <- binomial(link = logit)$linkinv(dat$eta[1] + dat$eta[2]*dat$x)
+dat$phi <- binomial(link = log)$linkinv(dat$gamma[1] + dat$gamma[2]*dat$z)
+dat$y <- rbeta(dat$N, dat$mu * dat$phi, (1 - dat$mu) * dat$phi)
```
jgabry: Same question as my question about this in the example for `stan_betareg.R`. Why the `dat` if it just means more work and code clutter calling `dat$` every time you want to do anything?

vignettes/betareg.Rmd
```diff
@@ -0,0 +1,195 @@
...
+```
+
+The usual posterior analyses are available in **rstanarm**.
+The plots below illustrate simulated values of the outcome variable. The incorrect
+model noticeably fails to capture the top of the distribution consistently in comparison
+to the true model.
+```{r simulated-analysis, fig.height=5}
+library(gridExtra)
+plot1 <- pp_check(fit1) + ggtitle("True Model: y ~ x | z")
+plot2 <- pp_check(fit2) + ggtitle("False Model: y ~ x")
```
jgabry: Are these still not looking right? If not then I need to spend some time figuring out what's up with that.

vignettes/betareg.Rmd
```diff
@@ -0,0 +1,195 @@
...
+library(loo)
+gas_loo1 <- loo(gas_fit1)
+gas_loo2 <- loo(gas_fit2)
+compare(gas_loo1, gas_loo2)
+```
+
+Additionally, evaluating the expected log predictive distribution using `loo`
+reveals that the second of the two models is preferred.
+
+Note that the `sqrt` link function (applied to the second set of regressors) is known to be unstable, so it
```
jgabry: Unless you're using `sqrt` link in the vignette I think you can get rid of this given that it's now in the function doc.

bgoodri:
We really need to get StanHeaders and rstanarm onto CRAN early this afternoon (both because StanHeaders is broken and because of the rstanarm webinar Tuesday). So, @imadmali if you have locally the fixes @jgabry suggested, then push them ASAP. Otherwise, we might have to hold off on `stan_betareg` for this release.

R/stan_betareg.fit.R
imadmali: Yep. And I changed it in stanreg.R as you suggested below.

R/stan_betareg.R
```diff
@@ -0,0 +1,189 @@
...
+  out <- poisson(link)
+  out$family <- "beta_phi"
+  out$variance <- function(mu, phi) mu * (1 - mu) / (phi + 1)
+  out$dev.resids <- function(y, mu, wt)
+    stop("'dev.resids' function should not be called")
+  out$aic <- function(y, n, mu, wt, dev)
+    stop("'aic' function should not have been called")
+  out$simulate <- function(object, nsim)
+    stop("'simulate' function should not have been called")
+  return(out)
```
imadmali: I removed the duplicate in 23a2648. Not sure why it didn't collapse here.

R/stan_betareg.fit.R
jgabry: Here I was actually referring to the object `is_beta` rather than the function `is.beta`.

imadmali: Gotcha. And no, it's an irrelevant check.

tests/testthat/test_stan_betareg.R
```diff
@@ -0,0 +1,127 @@
...
+                        prior = NULL, prior_intercept = NULL,
+                        prior_z = NULL, prior_intercept_z = NULL,
+                        data = dat, algorithm = "optimizing")
+    expect_stanreg(fit)
+    val <- coef(fit)
+    ans <- coef(betareg(y ~ x | z, link = link1[i], link.phi = link2[1], data = dat))
+    expect_equal(val, ans, tol = 0.1, info = c(link1[i], link2[1]))
+  }
+})
+
```
jgabry: If we are supporting `weights` and `offset` can you also add a quick test that just makes sure `stan_betareg` runs when `weights` and `offset` are specified and that the returned stanreg object has the correct  `weights` and `offset` values?

exec/continuous.stan
imadmali: @jgabry if you remove it then won't the one of the link functions fail (`log` I think)? I thought this is how you resolved that issue.

tests/testthat/test_stan_betareg.R
jgabry: this was intended as a separate comment not related to the comment about `.DS_store` files

exec/continuous.stan
jgabry: And I guess it could just be done inside the same loop that defines `mu` instead of a separate one

tests/testthat/test_stan_betareg.R
imadmali: I put in a basic test (using one link function). The errors that come up when using other link functions pertain to a beta_rng() issue which will be resolved in the next Stan release. After that release we can put in more comprehensive offset/weight tests. Is that ok?

exec/continuous.stan
imadmali: Sorry I wasn't clear... It resolves the log link issue we were having, but it applies to all mu link functions. (See top of p4 https://www.ime.usp.br/~sferrari/beta.pdf).

R/stan_betareg.fit.R
imadmali: I resolved this in cfcf260 and added adjusted the if statements that define `Z_true` accordingly. It follows the format of `betareg.fit` which defaults to `z = NULL`.

exec/continuous.stan
jgabry: The constraint applies to them all but some of them will automatically satisfy it. As is, this is checking for `<` or `>` and not `<=` or `>=`, so I think we know in advance that `mu[n] < 0 || mu[n] > 1` is `FALSE` whenever the inverse-link is a valid CDF.

R/stanreg.R
imadmali: I'm not sure why I wrote this. It seems that the way the residuals are done now match that of betareg.

```
dat <- list()
dat$N <- 200
dat$x <- rnorm(dat$N, 2, 1)
dat$z <- rnorm(dat$N, 2, 1)
dat$mu <- binomial(link = "logit")$linkinv(0.5 + 0.2*dat$x)
dat$phi <- exp(1.5 + 0.4*dat$z)
dat$y <- rbeta(dat$N, dat$mu * dat$phi, (1 - dat$mu) * dat$phi)
dat <- data.frame(dat$y, dat$x, dat$z)
colnames(dat) <- c("y", "x", "z")

check <- betareg::betareg(y ~ x | z, data = dat, link = "log", link.phi = "log")
fit <- rstanarm::stan_betareg(y ~ x | z , data = dat, link = "log", link.phi = "log",
                              chains = 4, cores = 4, algorithm = "sampling")

plot(fit$residuals)
points(check$residuals, col = "red")
```

bgoodri:
The rstanarm package got approved by CRAN, so you guys don't necessarily
have to sort all this out today. It would have been good to get betareg in
this time, but we (at best) barely have enough time for CRAN to build
binaries rstanarm packages for OSX and Windows before Tuesday.

tests/testthat/test_stan_betareg.R
```diff
@@ -125,3 +125,25 @@ test_that("stan_betareg returns expected result when modeling x and z using link
     expect_stanreg(fit)
   }
 })
+
+# test weights/offset (make test more comprehensive once the beta_rng() update is in stan math)
```
jgabry: test looks fine for now, thanks!

exec/continuous.stan
imadmali: Got it. I move it into the exp link loop in e61b467.

R/stan_betareg.fit.R
```diff
@@ -201,11 +201,11 @@ stan_betareg.fit <- function (x, y, z, weights = rep(1, NROW(x)), offset = rep(0
         show_messages = FALSE)
       stanfit <- do.call(sampling, sampling_args)
     }
-    else if (algorithm == "meanfield") { # FIXME
+    else if (algorithm == "meanfield") {
       stanfit <- rstan::vb(stanfit, pars = pars, data = standata,
                            algorithm = algorithm, init = 0.001, ...)
     }
-    else if (algorithm == "fullrank") { # FIXME
+    else if (algorithm == "fullrank") {
```
imadmali: Yep. Fixed in 099ef94.

jgabry:
Closes #124 when merged

R/stan_betareg.R
```diff
@@ -121,7 +121,12 @@ stan_betareg <- function (formula, data, subset, na.action, weights, offset,
   offset <- validate_offset(as.vector(model.offset(mf)), y = Y)
   if (!length(prior_ops)) 
     prior_ops <- list(scaled = FALSE, prior_scale_for_dispersion = Inf)
-
+  
+  # determine whether user specified regression matrix for precision model
+  if (length(grep("\\|", all.names(formula))) == 1 && length(link.phi) > 1) {
```
jgabry: Ok but now I'm confused why this is here but `match.arg(link)` and `match.arg(link.phi)` occur below _after_ fitting the model.  I think `link` and `link.phi` should be validated before passing them on to another function (`stan_betareg.fit`). And simply doing `link.phi <- match.arg(link.phi)` will take care of ensuring that `link.phi` has length 1 (it will be `"log"` if user leaves the default).

R/stan_betareg.fit.R
```diff
@@ -20,7 +20,7 @@
 
 stan_betareg.fit <- function (x, y, z = NULL, weights = rep(1, NROW(x)), offset = rep(0, NROW(x)),
                               link = c("logit", "probit", "cloglog", "cauchit", "log", "loglog"), 
-                              link.phi = c("log", "identity", "sqrt"), ...,
+                              link.phi = "log", ...,
```
jgabry: Again, I don't think you need to change the default arguments here. Just use `match.arg` below.

jgabry: Sorry if my comments from earlier today were confusing. I was trying to say that I don't think basing anything on the length of `link.phi` is the right way to go because that means the argument with  multiple options is treated differently than is typically expected.

imadmali: Your comments aren't confusing. It's just that betareg specifies the model in a peculiar way and I'm using differing link.phi lengths and "|" to capture that info and pass it on to `stan_betareg.fit`. For instance, I need to distinguish between,
```
betareg(y ~ x , data = dat, link = "log")
```
and
```
betareg(y ~ x , data = dat, link = "log", link.phi = "log")
betareg(y ~ x | 1, data = dat, link = "log")
betareg(y ~ x | 1, data = dat, link = "log", link.phi = "log")
```
So going back to what you said in a previous comment:

> ... it's only unnecessary but not wrong for a user to specify link.phi="log" or whatever single value even if there's no Z.

If there is no Z specified by the user then the type of model fit if link.phi is not specified by the user is different than if the user does specify link.phi. So,
```
betareg(y ~ x , data = dat, link = "log")
betareg(y ~ x , data = dat, link = "log", link.phi = "log")
```
are different models. 

---
But, given your comments about `match.arg`, in `stan_betareg` I can do 
```
 if (length(grep("\\|", all.names(formula))) == 1 && length(link.phi) > 1) {
    link.phi <- match.arg(link.phi)
  }
```
and in `stan_betareg.fit` I can do
```
  if (length(link.phi) > 1 || is.null(z)) {
    Z_true <- 0
    z <- model.matrix(y ~ 1)
    link.phi <- match.arg(link.phi)
  }
  else if (length(link.phi) == 1) {
    Z_true <- 1
  }
```

---
And I can pass the `match.arg` statements for `link` and `link.phi` in stan_betareg.fit before the model fitting happens (regardless though match.arg is called towards the beginning of stan_betareg.fit). 

Sorry that this explanation is so long-winded.

vignettes/betareg.Rmd
imadmali: No this was a user fault. I needed to specify an identical domain/range for the two plots.

R/stan_betareg.fit.R
jgabry: I'm going through and making a few tweaks now. I can take care of this while I'm at it

jgabry: Now that I look more closely at the betareg package, it looks like `stan_betareg` doesn't match`betareg` in terms of the default settings of `link.phi`. Is that intentional? I see these defaults for `link.phi`:

* `betareg`: `link.phi = NULL`
* `stan_betareg`: `link.phi = c("log", "identity", "sqrt")`

imadmali:
No. I was in the process of doing so but then I wasn't sure since you
suggested match.arg. I'll apply the betareg argument defaults to
stan_betareg. I think that actually makes it easier to capture the the
differences in models that I was talking about in a previous comment.

*IMAD ALI*

imadali.net <http://www.imadali.net/>

jgabry:
Ok cool. Thanks!

imadmali:
I've made the `stan_betareg` link defaults match the `betareg` defaults. I did the same for `stan_betareg.fit` except for the fact that `betareg.fit` doesn't allow users to model phi (precision) as a parameter (which I don't think is a good idea).

In the examples below, `betareg.fit` returns the same parameters whereas `stan_betareg.fit` will model phi as a parameter is z is undeclared.

```
# dgp
dat <- list()
dat$N <- 200
dat$x <- rnorm(dat$N, 2, 1)
dat$z <- rnorm(dat$N, 2, 1)
dat$mu <- binomial(link = "logit")$linkinv(0.5 + 0.2*dat$x)
# dat$mu <- dat$x*0.2
# dat$mu <- exp(dat$mu - max(dat$mu) + 0.5)
dat$phi <- exp(1.5 + 0.4*dat$z)
dat$y <- rbeta(dat$N, dat$mu * dat$phi, (1 - dat$mu) * dat$phi)
# hist(dat$y, col = "dark grey", border = F, xlim = c(0,1))
dat <- data.frame(dat$y, dat$x, dat$z)
colnames(dat) <- c("y", "x", "z")

# model fitting
check <- betareg::betareg.fit(x = model.matrix(y ~ x, data = dat),
                              y = dat$y)
fit <- rstanarm::stan_betareg.fit(x = model.matrix(y ~ x, data = dat),
                                  y = dat$y,
                                  link = "logit", algorithm = "sampling", iter = 1000, cores = 4)
coef(check)
fit

check <- betareg::betareg.fit(x = model.matrix(y ~ x, data = dat),
                              y = dat$y,
                              z = model.matrix(y ~ 1, data = dat))
fit <- rstanarm::stan_betareg.fit(x = model.matrix(y ~ x, data = dat),
                                  y = dat$y,
                                  z = model.matrix(y ~ 1, data = dat),
                                  link = "logit", algorithm = "sampling", iter = 1000, cores = 4)
coef(check)
fit
```

jgabry:
On Wed, Nov 30, 2016 at 12:19 AM, Imad Ali <notifications@github.com> wrote:

> I've made the stan_betareg link defaults match the betareg defaults. I
> did the same for stan_betareg.fit except for the fact that betareg.fit
> doesn't allow users to model phi (precision) as a parameter (which I don't
> think is a good idea).
>
> In the examples below, betareg.fit returns the same parameters whereas
> stan_betareg.fit will model phi as a parameter is z is undeclared.
>
Ok cool. As long as that difference between rstanarm and betareg is
documented then that seem ok to me.

bgoodri:
Why don't we think it is a good idea to model phi as a scalar parameter?

imadmali:
It is a good idea and we do it in stan_betareg() and stan_betareg.fit().
The user as the option to model phi as a parameter itself or to model phi
as a parameter times a vector of ones. betareg() also does this, but
betareg.fit() doesn't allow users to model phi as a parameter itself for
some reason. If you omit z it estimates y ~ x | 1 instead of y ~ x.

*IMAD ALI*

imadali.net <http://www.imadali.net/>

jgabry:
@imadmali I just noticed that while there are tests for `stan_betareg` itself, none of the other test files include tests using models fit with `stan_betareg`. For example, `test_posterior_predict.R`, `test_methods.R`, `test_loo.R`, etc.

R/stan_betareg.R
```diff
@@ -0,0 +1,198 @@
...
+  # pass the prior information to stan_betareg.fit()
+  stanfit <- stan_betareg.fit(x = X, y = Y, z = Z, 
+                              weights = weights, offset = offset, 
+                              link = link, link.phi = link.phi, ..., 
+                              prior = prior, prior_z = prior_z, 
+                              prior_intercept = prior_intercept, 
+                              prior_intercept_z = prior_intercept_z,
+                              prior_ops = prior_ops, prior_PD = prior_PD, 
+                              algorithm = algorithm, adapt_delta = adapt_delta,
+                              QR = QR, sparse = FALSE)
```
jgabry: @imadmali Also just wanted to check again about why `sparse` is fixed to `FALSE` here

imadmali: It shouldn't be. I'll fix it.

jgabry:
@imadmali @bgoodri Question about `stan_betareg` and priors. Are we not doing any of the rescaling of the priors for `stan_betareg` that we do for `stan_glm`? If so, I can't find that in the code (can you point me to where). If not, is that intentional?

imadmali:
@jgabry I thought the all the prior rescaling was taking place in `priors_glm.stan` and `priors_betareg.stan`. Where is the prior rescaling that you are referring to in `stan_glm`? I'll see if it can be implemented in `stan_betareg`.

jgabry:
@imadmali https://github.com/stan-dev/rstanarm/blob/feature/betareg/R/stan_glm.fit.R#L127-L146

jgabry:
I'm not sure if we want to do it or not for beta regression (haven't really
thought about it) but we currently do it for anything that calls
stan_glm.fit. The lines of code that I highlighted in my previous comment
correspond to what's described in the doc for the "scaled" argument to the
prior_options function.

bgoodri:
I think it makes sense to scale the prior on the coefficients according to
the standard deviation of the corresponding predictor. But I don't think we
want to also scale according to the standard deviation of the outcome
because we already have the phi parameter.

jgabry:
On another note, although it looks like travis is passing for this PR/branch it's actually timing out: 

https://api.travis-ci.org/jobs/182157200/log.txt?deansi=true

So it may or may not actually be passing R CMD check.

imadmali:
798fb4f scales the prior coefficients according to the predictors.
Regarding Travis: The tests pass when I run `devtools::check()` locally. It might be timing out because I've added some models that are using `sampling` in tests.

jgabry:
@imadmali I'm pretty sure `stan_betareg` and `stan_betareg.fit` will need some tweaking after the changes you just merged in from master

imadmali:
@jgabry from what I see the major changes are the ways in which the priors are specified and additional #includes in `continuous.stan`. Let me know if there's anything else. Also, although the latest commit errors on travis the tests pass locally.

R/stan_betareg.fit.R
```diff
@@ -24,7 +24,8 @@ stan_betareg.fit <- function(x, y, z = NULL,
                              link.phi = NULL, ...,
                              prior = normal(), prior_intercept = normal(),
                              prior_z = normal(), prior_intercept_z = normal(),
-                             prior_ops = prior_options(), prior_PD = FALSE, 
+                             prior_dispersion = cauchy(0, 5),
+                             prior_ops = NULL, prior_PD = FALSE,
```
jgabry: `prior_ops` should be removed entirely instead of changed to `prior_ops=NULL`. The only reason it hasn't been removed entirely from `stan_glm.fit` is to maintain backwards compatibility, but there's no backwards compatibility to maintain for `stan_betareg` because it's new. So we should just never introduce `prior_ops` at all.

R/stan_betareg.fit.R
```diff
@@ -110,14 +109,28 @@ stan_betareg.fit <- function(x, y, z = NULL,
...
+  
   if (nvars_z == 0) {
       prior_mean_z <- double()
       prior_scale_z <- double()
       prior_df_z <- integer()
   }
   
   # prior scaling (using sd of predictors)
-  if (scaled && prior_dist > 0L && !QR) {
+  min_prior_scale <- 1e-12 # used to be set in prior_options()
```
jgabry: can you remove the comment about `prior_options()`? That really only applies for `stan_glm.fit` because `stan_betareg.fit` hasn't been included in a release yet

R/stan_betareg.fit.R
```diff
@@ -110,14 +109,28 @@ stan_betareg.fit <- function(x, y, z = NULL,
...
+  if (prior_dist > 0L && !QR) {  # scaled &&
```
jgabry: I think the rescaling stuff should also depend on the new `autoscale` argument like it does for `stan_glm`. See e.g. https://github.com/stan-dev/rstanarm/blob/master/R/stan_glm.fit.R#L175, https://github.com/stan-dev/rstanarm/blob/master/R/data_block.R#L94

jgabry:
@imadmali I just made a few more comments but aside from I mentioned in those comments I think it looks ok

R/stan_betareg.fit.R
imadmali: Are you referring to `prior_ops` not being user defined or not being in `stan_betareg.fit` at all. If I pull out prior_ops = NULL then I have to pull it out of `prior_info` here https://github.com/stan-dev/rstanarm/blob/feature/betareg/R/stan_betareg.fit.R#L216 and from `summarize_betareg_prior` here https://github.com/stan-dev/rstanarm/blob/feature/betareg/R/stan_betareg.fit.R#L302 which causes some of your prior_summary stuff to break.

jgabry: I meant not in `stan_betareg.fit` at all. `summarize_betareg_prior` will just need to be tweaked to get that information without needing `prior_ops`, similarly to how we changed  `summarize_glm_prior` to not require `prior_ops`.

imadmali: Got it. d275d45 has `prior_ops` removed and adjusts `summarize_betareg_prior` accordingly. (Also had to update some tests.)

imadmali:
Ah. I need to catch these before I commit. Sorry.

jgabry:
@imadmali Do you think this is ready to go or are things still things you were intending to tweak?

jgabry:
@imadmali Is `prior_dispersion` used for anything in models `z` variables? Or is it just ignored?

imadmali:
I don't have any additional changes to make unless you find something.
`prior_dispersion` is ignored when phi is modeled with `z` variables. It should only be used when modeling `phi` as a parameter (i.e. no `z` variables).

jgabry:
Ok is that documented anywhere? Also it would be good if prior_summary and
things like that didn't include the dispersion when it's ignored. Not sure
what the current behavior is.

imadmali:
I'll mention it in the documentation and adjust prior summary accordingly.

jgabry:
Thanks. Actually, peeking at it now, I'm not sure the dispersion is ever included, even for models that use it.

imadmali:
The `dispersion` parameter is being used here:
https://github.com/stan-dev/rstanarm/blob/feature/betareg/exec/continuous.stan#L127-L131
The priors for `dispersion` are here:
https://github.com/stan-dev/rstanarm/blob/feature/betareg/exec/continuous.stan#L152-L165
At the moment, `prior_dispersion` is mapping to `prior_..._for_dispersion` which is being passed into `continuous.stan`. Should it be done in a different way?

jgabry:
@imadmali While you're fiddling with the dispersion stuff, it doesn't actually make sense to call this a dispersion parameter, since it's not a dispersion parameter, but rather the reciprocal of something kind of like a dispersion parameter (but not quite). In the comments to PR #147 @bgoodri suggests:

> So, let's call it `phi`, say somewhere that it is the same likelihood as in betareg, and ignore the fact that the betareg doc also calls it a precision parameter even though it is not a reciprocal variance. The summary.betareg method doesn't seem to say anything but `phi` anywhere.

jgabry:
@imadmali I meant that it wasn't ever included in `prior_summary`

imadmali:
Yeah I see that now. 
I've been following PR #147 but I'm a little confused... is the `phi` stuff something that we just want to change in the documentation/vignette or do we want a new parameter `phi` declared in `continuous.stan`.

jgabry:
The change would be in the doc and R code but not the Stan code. #147 changes `dispersion` to `aux` in `continuous.stan`. Should be the same for beta regression. So you can do everything in terms of `phi` in R and then call it `aux` when passing to Stan.

imadmali:
Aside from declaring prior stuff, in `stan_betareg` we don't actually do anything with `dispersion` before model fitting; everything is done in `continuous.stan`.  So with regard to `dispersion` itself there's nothing to change in terms of passing anything to Stan. 
Does `prior_dispersion` and `prior_..._for_dispersion` need to change to `prior_aux` and `prior_..._for_aux`?

jgabry:
I think the simplest way to think about it is that anything user-facing should use `phi` and anything internal `aux`.

imadmali:
Okay, in which case I'll have the user declare `prior_phi = ...` (instead of `prior_dispersion = ...`) in `stan_betareg()` and `stan_betareg.fit()`. (Just like they have to declare `link.phi = ...` when relevant.)

bgoodri:
Yes

imadmali:
Okay so for any user-facing stuff I replaced dispersion with phi. But I'm gonna hold off on replacing dispersion with aux within `stan_betareg.fit()` until PR #147 has merged as @jgabry mentioned.

imadmali:
Yeah that's more efficient. I'll make the change.

imadmali:
Okay it looks like PR #147 has been merged so I'll make the appropriate changes to the feature/betareg branch.

jgabry:
Cool. I already merged from master and fixed the conflicts.

imadmali:
Thanks, it sped things up.

jgabry:
@bgoodri I think stan_betareg is probably good to merge

PR #138, opened 2016-12-09
Replace prior_options

jgabry:
This PR deprecates the `prior_options` function (used to specify `prior_ops` argument) and implements the following functionality to replace and expand the functionality of `prior_options`: 

* Introduces the `prior_dispersion` argument to all modeling functions that call `stan_glm.fit`. `prior_dispersion` can be a call to `normal`, `student_t`, `cauchy`, or `exponential`.  The default for `prior_dispersion` is `cauchy(0, 5)`, which maintains backwards compatibility (should we change this default?). 
* Introduces the `exponential` function for use as a prior on scale/dispersion parameters
* Adds argument `autoscale` to functions `normal`, `student_t`, and `cauchy`. Can be set to `FALSE` to turn off rstanarm’s internal adjustments to scale parameters. This replaces `prior_options(scaled = FALSE)`.


jgabry:
Just need to add something like

#' @param prior_ops Deprecated. See \link{rstanarm-deprecated} for details.

to stan_glm.fit.R.

I'm not at my computer but I can do that late tonight. Or feel free to add
that if you want to merge it sooner.

jgabry:
@bgoodri I think this passes now. If you haven't already, can you double check the Stan code before merging?

exec/continuous.stan
```diff
@@ -234,9 +234,15 @@ parameters {
 transformed parameters {
   real dispersion;
   #include "tparameters_glm.stan" // defines beta, b, theta_L
-  if (prior_scale_for_dispersion > 0)
-    dispersion =  prior_scale_for_dispersion * dispersion_unscaled;
-  else dispersion = dispersion_unscaled;
+  
+  if (prior_dist_for_dispersion == 0 || prior_scale_for_dispersion <= 0)
```
bgoodri: Is it possible for `prior_scale_for_dispersion` to be `<= 0` when `prior_dist_for_dispersion > 0`?

PR #139, opened 2016-12-14
issue 47 loo checks

### LINKED ISSUE #47: Check that y is the same for model comparison (loo or waic)
The `compare` function in the loo package checks that models have the same number of observations, but we can also check that the outcome variable is the same. This would probably require a wrapper around `loo::compare` that first does this check, which means rstanarm would have its own model comparison function, instead of users just using the loo package. But I'm not sure that doing this check is worth adding a function since `compare`'s check for the number of observations will catch almost all cases of y being different.

jgabry:
This PR introduces `compare_models`, which is a wrapper around `loo::compare` that also checks that the __rstanarm__ models being compared have the same outcome variable and the same family/link. 

@bgoodri @avehtari  To check that outcome variables are the same I first use the `sha1` function from the __digest__ package to calculate a SHA1 hash of the outcome (the hash is then stored as an attribute in the loo object so it is accessible at the time of model comparison). Is there a better way to do this? Or a different hashing algorithm I should use? This also means we need the __digest__ package in `Suggests` now.

closes #47


bgoodri:
What is wrong with using the `identical()` function to check whether to outcome vectors are the same? Also, I don't think it is required that the models have the same link function, only the same likelihood.

jgabry:
They only get checked once compare_models is called so we have the loo
objects and not the stanreg objects to work with. So to use identical we
would need to save the whole outcome variable in the loo object so they
could be compared later. That could conceivably be large. With the hash we
only need to add a shortish string to the loo object and then we can
compare the hashes.

And yeah, you're right about the link. I should just be checking family
names not family objects.

bgoodri:
I guess since digest is used by a lot of packages we are already importing
from or suggesting, it is not a big deal to use it for loo().

jgabry:
I couldn't find a way to do it using base or utils, and the hashing in
tools looks like it's for files not objects. Do you know of a way to hash
something using one of the code R packages?

It would be simpler to use identical, and most outcomes aren't too big so
we could do that. But we don't actually need the outcome values, just that
they are the same. Actually, might identical be too strict in some cases?

jgabry:
I meant "core" not "code" R packages

avehtari:
I assume you allow comparing models which may have different families as long as it's continuous vs continuous or count vs count?

jgabry:
@avehtari So, for example, should we be allowing model with Gamma likelihood to be compared to model with Gaussian likelihood?

avehtari:
Oops, I thought I had asnwered this already.

Gamma and inverse Gaussian are different from Gaussian as they have different support, but it would be natural to compare Gamma and inverse Gaussian likelihoods.

For count data it's very common to compare Poisson and negative Binomial distributions. Sometimes binomial model is approximated with Poisson.

So I would only check that in the comparison continuous and discrete distributions are not mixed (it's possible that someone would use continuous distribution to model discrete data, but then when making the comparison vs discrete distribution it would be necessary to compute probabilities for intervals of continuous distribution to make them comparable).

Aki

jgabry:
Interesting. I certainly see why comparing e.g. poisson to NB is appealing,
but I have for the most part been working under the assumption that the
likelihood functions should have the same general form. So one could
perhaps justify comparing NB to poisson with noise terms added to linear
predictor but not to plain poisson. I think @bgoodri was also thinking
that. From what you write I guess this PR is too strict?

avehtari:
Yes, I think it is too strict. It is common that people want to compare different observation models.
I think there should be checks only for
1) y is same for all compared models
    - This will prevent people making a mistake by comparing, e.g., y~gamma() vs. log(y)~normal. This comparison is valid if taking into account the jacobian of transformation of y, but rstanarm can't know which transformation user used.
2) observation models (families?) have to be all continuous or all discrete
   - This will prevent people making a mistake comparing, e.g., y~poisson() vs. y~normal. This comparison can be made valid by computing probabilities of intervals in case of continuous distribution, but it's low priority feature.

jgabry:
Ok I just pushed a change so it checks for discrete vs continuous instead of likelihood function. It doesn't check for common support so technically Gaussian vs gamma is allowed because both are continuous. It would be slightly more tedious to check this but it's doable if it's important.

jgabry:
@imadmali I'm adding you as a "reviewer" on some recent PR's to indicate they may be relevant to whatever you have going on with the beta regression stuff

bgoodri:
This is ready right?

jgabry:
I believe it was ready last time I checked. I don't think there's been anything merged recently that would affect this, right?

Issue #114, opened 2016-09-05
Feature: more flexibility in specifying global scale in the HS-prior


jpiironen:
#### Summary:

I would like a possibility of adjusting the scale and degrees of freedom for the global parameter tau in the hierarchical shrinkage (HS) prior.
#### Description:

We discussed the use of HS-prior for regression models in our technical report (Piironen and Vehtari, 2015) after which it was added to rstanarm, which is great! However, the current version allows only tuning the degrees of freedom for the local scale parameters lambda_j, but does not allow changing the prior for the global parameter tau which is given half-Cauchy prior with scale 1. For some datasets, this prior for tau gives too much mass for implausibly large values, and therefore favors solutions with many of the regression coefficients different from zero (I'm working on a paper that would give a recommendation on how to specify prior for tau based on our guess for the number of relevant variables in the data). A suggestion:

1) Replace tau ~ half-Cauchy(0,1) with tau ~ half-t_nu(0,s), and allow user to change the degrees of freedom nu and scale s (by default they can be set to 1 and 1, however). 

2) In regression, there are theoretical considerations why one might want to use prior tau ~ half-t_nu(0,s*sigma), where sigma is the noise std. So it would be nice to have this alternative too. 

I believe item 1) should be rather easy and quick to implement but would be very useful by providing more control over tau. However, item 2) might require more extra work and there might be some practical difficulties in the implementation and use I'm not aware of. Thus I'm proposing it only as something you could consider in the future, and I think item 1) would be more important to have at the moment.

Thanks in advance!
#### References:

J. Piironen, A. Vehtari (2015). Projection predictive variable selection using Stan+R, http://arxiv.org/abs/1508.02502

jgabry:
Hi Juho, thanks for the suggestion. It seems like a good idea to me.
@bgoodri what do you think?

For (1) we could preserve backwards compatibility if we add a second
argument to hs, something like hs(df=3, global_scale = 1).

For (2) I agree that some more research is probably needed, but the coding
shouldn't be too bad.

bgoodri:
Yeah

jgabry:
@bgoodri looks like you reparameterized this in

https://github.com/stan-dev/rstanarm/blob/master/inst/chunks/priors_glm.stan

at some point. I'm sure I could figure it out but do you know off the top
of your head what I'd need to tweak there to set there that would equate to
changing the half-Cauchy scale parameter?

bgoodri:
[...]

PR #140, opened 2016-12-14
Feature more flexible hs prior

bgoodri:
not quite ready to merge yet but let's make a PR


bgoodri:
@jpiironen @avehtari Is this all you need for the time being?

jpiironen:
Seems great! I realized you implemented at least the first feature that I proposed in the issue (https://github.com/stan-dev/rstanarm/issues/114), but I wasn't sure if you implemented also the second alternative (where the scale of tau depends on the unknown regression noise std sigma)? It may also be my mistake that I did not find it, but it would be useful to have that too. There is also an example Stan implementation of this prior in the appendix of our arXiv paper (https://arxiv.org/abs/1610.05559, page 11) if you want to have a look. Thanks.

bgoodri:
I think the second thing about sigma will have to wait. We don't have an obvious way to specify options like that.

avehtari:
Oh, that's unfortunate. Some week ago we learned from James Scott, one of the horseshoe prior paper authors, that they also would always use C+(0, sigma) instead of C+(0,1) and during the discussion he realized that there's a mistake in the paper stating  C+(0,1) instead of C+(0, sigma) used for the experiments. So it seems that the sigma should be there by default, and there is no need for an option to have it otherwise.

bgoodri:
That would be easier.

bgoodri:
It is straightforward to scale by sigma in Gaussian models but is that the only case where we should scale a HS (plus) prior? What about Gamma and inverse Gaussian likelihoods? What about negative binomials?

bgoodri:
Also, would people also recommend scaling a Laplace prior in some cases but not others?

avehtari:
It would be natural to scale with all models. In https://arxiv.org/abs/1610.05559 we have corresponding scaling for Bernoulli model and @jpiironen will write a note about scaling for other models. @jpiironen thinks that even if scaling with the scale of likelihood is added, there should be a way to provide the guess for the sparseness.

bgoodri:
@jgabry Maybe we should always scale by the error standard deviation and then not scale up the prior standard deviations by the standard deviation of y?

jgabry:
We're talking about the scale of the prior for tau in hs and hs_plus right?
If so then I think to always do that is fine.

Or did you means always use sigma rather than sd(y) for all glms with
adjusted prior scales?

bgoodri:
Both, maybe

jgabry:
For the former I think it's safe to go ahead and do that now. For the
latter, I think we'd probably want to do some comparisons first. What would
we gain from making the switch? We'd be on more theoretically solid ground,
I guess, but do you think it would improve estimates in practice?

jpiironen:
Just to clarify, my idea was that the prior scale for tau would be b*s, where b would be given by the user and s = sqrt(-1/L_i''), where L_i'' denotes the second derivative of the ith log-likelihood term w.r.t. the linear predictor f = x*beta. For Gaussian likelihood this gives s = sigma. For other likelihoods s is generally not even a constant, but it seems reasonable plug-in values could be derived (for instance s = 2 for Bernoulli likelihood with logistic link).

At the moment I'm deriving results for some other likelihoods, but doing this for all combinations of the observation model and link function seems cumbersome and slows down adding this functionality for those cases where we know exactly what to do (e.g. Gaussian). My other concern is that for some of the cases selecting the plug-in value is not a trivial choice, and if these are hard-coded in rstanarm and later we realize a different choice might have been better, we need to either change the code or live with the earlier choice. Any thoughts on these?

bgoodri:
I think this is ready to merge. I put the new scaling things into the `hs_plus` prior.

bgoodri:
I spoke too soon

PR #143, opened 2016-12-18
Feature/laplace prior

bgoodri:
will merge as soon as merge conflicts are resolved


jgabry:
@bgoodri I just updated those doc pages I mentioned in the above comments. Anything else need to be done for this PR?

jgabry:
Do you want to also allow the lasso and laplace priors for `stan_betareg`?

jgabry:
I just went ahead and enabled them for `stan_betareg` also.

jgabry:
@bgoodri Is this PR ready to go? It looks ok to me, although why do you think we should include `product_normal`? Do you have a particular use case in mind for which it's important?

bgoodri:
It is almost ready. I don't think product_normal is that good a prior, but
it is good for illustrating why it is worse than it seems.

PR #146, opened 2016-12-23
Dispersion to nuisance

jgabry:
* Renames `dispersion` to `nuisance` pretty much everywhere it's used
* Renames `overdispersion` to `reciprocal_dispersion` in output from negative binomial models. 

"nuisance" is not an ideal term but currently it's "dispersion" regardless of whether that makes sense for the given model. "nuisance"  is less misleading.


jgabry:
fixed merge conflicts

bgoodri:
Possible other names: aux{illiary}, extra, other

jgabry:
How about we use auxiliary in text and aux in code?

bgoodri:
Fine by me

jgabry:
going to rename this branch now that we're using aux instead of nuisance. i'll open a new pull request

PR #147, opened 2016-12-30
Dispersion to aux

jgabry:
* Renames `dispersion` to `aux`
* Renames `overdispersion` to `reciprocal_dispersion` in output from negative binomial models.


jgabry:
@bgoodri It occurs to me another way to do this is to use `prior_aux` for functions like `stan_glm` and `stan_glmer` (for which multiple families are allowed), but use a family-specific name if there is a wrapper that supports only one family (e.g., `stan_lmer, stan_glm.nb, stan_glmer.nb`). 

For example, currently on this branch these calls estimate the same model: 

```r
stan_glmer(..., family = "gaussian", prior_aux = exponential(1)) 
stan_lmer(..., prior_aux = exponential(1)) 
```

but we could make it so that the second call is 

```r
stan_lmer(..., prior_sigma = exponential(1))  # or prior_error_sd, or whatever
```
where `prior_sigma` gets passed to `stan_glmer`'s `prior_aux` argument. 

This is also relevant for `stan_betareg`, for which `prior_dispersion` currently controls the prior on `phi` when there are no `z` variable specified. To match the rest of the package I guess that should be `prior_aux` now, but since `stan_betareg` only does beta regression it could make sense to leave it as `prior_dispersion`. But if we do that then for consistency we should probably also do it for other functions that use `continuous.stan` for a single family.

bgoodri:
On Fri, Dec 30, 2016 at 3:38 PM, Jonah Gabry <notifications@github.com>
wrote:

> @bgoodri <https://github.com/bgoodri> It occurs to me another way to do
> this is to use prior_aux for functions like stan_glm and stan_glmer (for
> which multiple families are allowed), but use a family-specific name if
> there is a wrapper that supports only one family (e.g., stan_lmer,
> stan_glm.nb, stan_glmer.nb).
>
> For example, currently on this branch these calls estimate the same model:
>
> stan_glmer(..., family = "gaussian", prior_aux = exponential(1))
> stan_lmer(..., prior_aux = exponential(1))
>
> but we could make it so that the second call is
>
> stan_lmer(..., prior_sigma = exponential(1))  # or prior_error_sd, or whatever
>
> where prior_sigma gets passed to stan_glmer's prior_aux argument.
>
I don't like it for stan_lmer, stan_glm.nb, and stan_glmer.nb because it
creates two different ways of specifying the same model. For things like
stan_betareg, it makes more sense to have something instead of prior_aux,
but the non-mean parameter in a beta distribution isn't a dispersion
parameter either. The variance goes down when it goes up. But it is not
strictly a reciprocal dispersion parameter either. Wikipedia calls it both
a "sample size" and a "concentration" parameter, neither of which are
really correct terminology.

jgabry:
On Fri, Dec 30, 2016 at 4:03 PM, bgoodri <notifications@github.com> wrote:

> On Fri, Dec 30, 2016 at 3:38 PM, Jonah Gabry <notifications@github.com>
> wrote:
>
> > @bgoodri <https://github.com/bgoodri> It occurs to me another way to do
> > this is to use prior_aux for functions like stan_glm and stan_glmer (for
> > which multiple families are allowed), but use a family-specific name if
> > there is a wrapper that supports only one family (e.g., stan_lmer,
> > stan_glm.nb, stan_glmer.nb).
> >
> > For example, currently on this branch these calls estimate the same
> model:
> >
> > stan_glmer(..., family = "gaussian", prior_aux = exponential(1))
> > stan_lmer(..., prior_aux = exponential(1))
> >
> > but we could make it so that the second call is
> >
> > stan_lmer(..., prior_sigma = exponential(1)) # or prior_error_sd, or
> whatever
> >
> > where prior_sigma gets passed to stan_glmer's prior_aux argument.
> >
> I don't like it for stan_lmer, stan_glm.nb, and stan_glmer.nb because it
> creates two different ways of specifying the same model. For things like
> stan_betareg, it makes more sense to have something instead of prior_aux,
> but the non-mean parameter in a beta distribution isn't a dispersion
> parameter either. The variance goes down when it goes up. But it is not
> strictly a reciprocal dispersion parameter either. Wikipedia calls it both
> a "sample size" and a "concentration" parameter, neither of which are
> really correct terminology.
>

Ok, well for `stan_betareg` we need to call it something and it's going to
be called dispersion unless we change it (since it is now). betareg calls
it precision, which also isn't exactly right, but is better than
dispersion. I think we should not call it dispersion, so the issue is then
which of the imperfect names (precision, reciprocal_dispersion,
sample_size, concentration) do we prefer? I guess I prefer precision to
keep with betareg, but my preference isn't super strong so if you have a
different one we can go with that. We should definitely decide before
merging Imad's PR, though.

bgoodri:
betareg() seems to call it phi

jgabry:
Yeah, but in the doc they refer to it as the "precision parameter", e.g. from `?betareg`

> Fit beta regression models for rates and proportions via maximum likelihood using a parametrization with mean (depending through a link function on the covariates) and **precision parameter (called phi).**

bgoodri:
It is referred to as phi in args(betareg). So, let's call it phi, say
somewhere that it is the same likelihood as in betareg, and ignore the fact
that the betareg doc also calls it a precision parameter even though it is
not a reciprocal variance. The summary.betareg method doesn't seem to say
anything but phi anywhere.

jgabry:
Ok that's fine by me.

jgabry:
@bgoodri we ready to merge this one?

bgoodri:
I'll look at it after I eat.

jgabry:
Ok cool. Merging this one will make some other things simpler (e.g. checking betareg branch)

PR #152, opened 2016-12-31
Betareg fixes

jgabry:
@imadmali We merged your PR but I noticed more things that need to fixed with `stan_betareg` so I’m starting this follow up PR.

### DIFF
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -47,11 +47,11 @@
 #'   "cloglog", "cauchit", "log", and "loglog" are supported.
 #' @param link.phi If applicable, character specification of the link function 
 #'   used in the model for \code{phi} (specified through \code{z}). Currently, 
-#'   "identity", "log" (default), and "sqrt" are supported. Since the "sqrt" link function
-#'   is known to be unstable, it is advisable to specify a different link 
-#'   function (or to model \code{phi} as a scalar parameter instead of via a 
-#'   linear predictor by excluding \code{z} from the \code{formula} and
-#'   excluding \code{link.phi}).
+#'   "identity", "log" (default), and "sqrt" are supported. Since the "sqrt"
+#'   link function is known to be unstable, it is advisable to specify a
+#'   different link function (or to model \code{phi} as a scalar parameter
+#'   instead of via a linear predictor by excluding \code{z} from the
+#'   \code{formula} and excluding \code{link.phi}).
 #' @param prior_z Prior distribution for the coefficients in the model for 
 #'   \code{phi} (if applicable). Same options as for \code{prior}.
 #' @param prior_intercept_z Prior distribution for the intercept in the model 
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -47,7 +47,7 @@
 #'   "cloglog", "cauchit", "log", and "loglog" are supported.
 #' @param link.phi If applicable, character specification of the link function 
 #'   used in the model for \code{phi} (specified through \code{z}). Currently, 
-#'   "identity", "log", and "sqrt" are supported. Since the "sqrt" link function
+#'   "identity", "log" (default), and "sqrt" are supported. Since the "sqrt" link function
 #'   is known to be unstable, it is advisable to specify a different link 
 #'   function (or to model \code{phi} as a scalar parameter instead of via a 
 #'   linear predictor by excluding \code{z} from the \code{formula} and
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -104,7 +104,6 @@
 #' 
 stan_betareg <- function(formula, data, subset, na.action, weights, offset,
                          link = c("logit", "probit", "cloglog", "cauchit", "log", "loglog"),
-                         # link.phi = c("log", "identity", "sqrt"), 
                          link.phi = NULL,
                          model = TRUE, y = TRUE, x = FALSE, ...,
                          prior = normal(), prior_intercept = normal(),
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -41,7 +41,6 @@
 #' @template args-algorithm
 #' @template args-adapt_delta
 #' @template args-QR
-#' @template args-sparse
 #' 
 #' @param link Character specification of the link function used in the model 
 #'   for mu (specified through \code{x}). Currently, "logit", "probit",
@@ -112,7 +111,7 @@ stan_betareg <- function(formula, data, subset, na.action, weights, offset,
                          prior_z = normal(), prior_intercept_z = normal(),
                          prior_phi = cauchy(0, 5), prior_PD = FALSE, 
                          algorithm = c("sampling", "optimizing", "meanfield", "fullrank"),
-                         adapt_delta = NULL, QR = FALSE, sparse = FALSE) {
+                         adapt_delta = NULL, QR = FALSE) {
   
   if (!requireNamespace("betareg", quietly = TRUE)) 
     stop("Please install the betareg package before using 'stan_betareg'.")
@@ -150,7 +149,7 @@ stan_betareg <- function(formula, data, subset, na.action, weights, offset,
                               prior_intercept_z = prior_intercept_z,
                               prior_phi = prior_phi, prior_PD = prior_PD, 
                               algorithm = algorithm, adapt_delta = adapt_delta,
-                              QR = QR, sparse = sparse)
+                              QR = QR)
   algorithm <- match.arg(algorithm)
   link <- match.arg(link)
   link_phi <- match.arg(link.phi, c(NULL, "log", "identity", "sqrt"))
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -45,13 +45,13 @@
 #' 
 #' @param link Character specification of the link function used in the model 
 #'   for mu (specified through \code{x}). Currently, "logit", "probit",
-#'   "cloglog", "cauchit", "log", "loglog" are supported.
+#'   "cloglog", "cauchit", "log", and "loglog" are supported.
 #' @param link.phi If applicable, character specification of the link function 
 #'   used in the model for \code{phi} (specified through \code{z}). Currently, 
 #'   "identity", "log", and "sqrt" are supported. Since the "sqrt" link function
 #'   is known to be unstable, it is advisable to specify a different link 
 #'   function (or to model \code{phi} as a scalar parameter instead of via a
-#'   linear predictor).
+#'   linear predictor by excluding \code{z} from the \code{formula} and excluding \code{link.phi}).
 #' @param prior_z Prior distribution for the coefficients in the model for 
 #'   \code{phi} (if applicable). Same options as for \code{prior}.
 #' @param prior_intercept_z Prior distribution for the intercept in the model 
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -50,8 +50,9 @@
 #'   used in the model for \code{phi} (specified through \code{z}). Currently, 
 #'   "identity", "log", and "sqrt" are supported. Since the "sqrt" link function
 #'   is known to be unstable, it is advisable to specify a different link 
-#'   function (or to model \code{phi} as a scalar parameter instead of via a
-#'   linear predictor by excluding \code{z} from the \code{formula} and excluding \code{link.phi}).
+#'   function (or to model \code{phi} as a scalar parameter instead of via a 
+#'   linear predictor by excluding \code{z} from the \code{formula} and
+#'   excluding \code{link.phi}).
 #' @param prior_z Prior distribution for the coefficients in the model for 
 #'   \code{phi} (if applicable). Same options as for \code{prior}.
 #' @param prior_intercept_z Prior distribution for the intercept in the model 
@@ -83,7 +84,6 @@
 #'   31(7), 799--815.
 #' 
 #' @examples 
-#' \donttest{
 #' ### Simulated data
 #' N <- 200
 #' x <- rnorm(N, 2, 1)
@@ -94,6 +94,7 @@
 #' hist(y, col = "dark grey", border = FALSE, xlim = c(0,1))
 #' fake_dat <- data.frame(y, x, z)
 #' 
+#' \donttest{
 #' fit <- stan_betareg(y ~ x | z, data = fake_dat, 
 #'                     link = "logit", link.phi = "log")
 #' print(fit, digits = 2)
--- a/R/stan_betareg.R
+++ b/R/stan_betareg.R
@@ -17,8 +17,8 @@
 
 #' Bayesian beta regression models via Stan
 #'
-#' Beta regression modeling with optional prior distributions for 
-#' the coefficients, intercept, and phi (when modeled as a parameter).
+#' Beta regression modeling with optional prior distributions for the 
+#' coefficients, intercept, and auxiliary parameter \code{phi} (if applicable).
 #'
 #' @export
 #' @templateVar armRef (Ch. 3-6)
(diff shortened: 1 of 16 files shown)


R/stan_betareg.R
```diff
@@ -43,33 +43,38 @@
...
-#' @param prior_phi See \code{prior_dispersion}.
+#' @param link Character specification of the link function used in the model 
+#'   for mu (specified through \code{x}). Currently, "logit", "probit",
+#'   "cloglog", "cauchit", "log", "loglog" are supported.
+#' @param link.phi If applicable, character specification of the link function 
+#'   used in the model for \code{phi} (specified through \code{z}). Currently, 
+#'   "identity", "log", and "sqrt" are supported. Since the "sqrt" link function
+#'   is known to be unstable, it is advisable to specify a different link 
+#'   function (or to model \code{phi} as a scalar parameter instead of via a
+#'   linear predictor).
```
jgabry: @imadmali I tweaked this but can you also add something here to clarify the part in parentheses at the end? That is, how does one not model phi as a linear predictor? I assume by not specifying a `z` part of the formula, but it would be good to make that explicit.

jgabry:
@imadmali Also, `stan_betareg` accepts the `QR` and `sparse` arguments but I'm not sure they work. Are you doing anything with the `QR` and `sparse` arguments anywhere? I don't see it in `stan_betareg.fit`. That is, it doesn't look like you actually do a QR decomposition anywhere if `QR` is `TRUE` and it doesn't look like you do any sparse matrix prep if `sparse` is `TRUE`. Am I wrong?

R/stan_betareg.R
imadmali: You model phi as a linear predictor by specifying `z`. If you don't specify `z` or `link.phi = ...` then phi will be modeled as a parameter. I'll clarify in the doc.

imadmali:
No, thanks for catching that. I've added a QR decomposition for `x` the way it is done in `stan_glm.fit`. I'm assuming we also want to do this with the `z` variables?

Is the `sparse` argument just a matter of not centering `x` and `z` when `sparse = TRUE`? In which case it is included in `stan_betareg.fit` here,
https://github.com/stan-dev/rstanarm/blob/betareg-fixes/R/stan_betareg.fit.R#L71-#L80
or is there more to it?

bgoodri:
There is more to the sparse option in stan_glm.fit, but it is not very
good. At the moment, it is much slower so you are only saving RAM. And I
doubt people doing beta regressions are RAM constrained.

jgabry:
@imadmali  I'd say get rid of `sparse` from `stan_betareg`, at least for now. 
@bgoodri any reason why we wouldn't also want to do the QR thing with the `z` variables?

bgoodri:
Only reason would be if there is only 1 z variable. But I wouldn't make a
QR_z argument for it.

jgabry:
I was thinking `QR=TRUE` would just mean it would do it for both

imadmali:
Okay, I'm in the process of putting in the QR stuff for `z`, and yeah `QR = TRUE` will only work if both `x` and `z` have more than one variable.

bgoodri:
If someone does QR = TRUE, it should apply to any x or z that have more
than one variable.

jgabry:
@imadmali is the default for `link.phi` documented somewhere? It's set to default to `NULL` but that's not really what it defaults to internally (seems to default to `"log"`).

tests/testthat/test_posterior_predict.R
```diff
@@ -253,13 +253,21 @@ test_that("compatible with stan_lmer with offset", {
...
-test_that("compatible betareg", {
+test_that("compatible with stan_betareg with z", {
   data("GasolineYield", package = "betareg")
-  fit <- SW(stan_betareg(yield ~ temp, data = GasolineYield, 
-                     iter = ITER*5, chains = CHAINS, seed = SEED, refresh = REFRESH))
+  fit <- SW(stan_betareg(yield ~ batch + temp | temp, data = GasolineYield,
+                         iter = ITER, chains = CHAINS, seed = SEED, 
+                         refresh = REFRESH))
   check_for_error(fit)
   expect_linpred_equal(fit)
```
jgabry: @imadmali this test I added to `test_posterior_predict` for `stan_betareg` fails. I added a test for a model including a `z` term because previously there was only one without it. It's just the `expect_linpred_equal` test that fails. What does `posterior_linpred` do for `stan_betareg` models that have `z` terms?

vignettes/betareg.Rmd
```diff
@@ -23,36 +23,40 @@ params:
...
+are $\alpha$ and $\beta$. The shape parameters enter into the model according to
+the following transformations.
 $$
 \alpha = \mu\cdot\phi \\
 \beta = (1-\mu)\cdot\phi
 $$
 
-Where $g_1(\mu) = \mathbf{X}\boldsymbol{\eta}$ where $g_1(\cdot)$ is some link function, $X$ is a $n\times K$ dimensional matrix of
-predictors, and $\eta$ is a $K$ dimensional vector of parameters associated with each predictor. 
+Where $g_1(\mu) = \mathbf{X}\boldsymbol{\eta}$ where $g_1(\cdot)$ is some link
```
jgabry: @imadmali I don't think you've introduced `g_1(mu)` yet so this seems out of place.

vignettes/betareg.Rmd
```diff
@@ -23,36 +23,40 @@ params:
...
+function, $X$ is a $n\times K$ dimensional matrix of predictors, and $\eta$ is a
+$K$ dimensional vector of parameters associated with each predictor.
```
jgabry: @imadmali  we typically use `eta` for the linear predictor and `beta` for the coefficients on `X`

vignettes/betareg.Rmd
```diff
@@ -23,36 +23,40 @@ params:
...
-the product of beta distributions.
+Steps 3 and 4 are covered in more depth by the vignette entitled ["How to Use
+the __rstanarm__ Package"](rstanarm.html). This vignette focuses on Step 1 when
+the likelihood is the product of beta distributions.
 
 # Likelihood
 
 Beta regression uses the beta distribution as the likelihood for the data,
 $$
 f(y_i | \alpha,\beta) = \frac{x_i^{(\alpha-1)}(1-x_i)^{(\beta-1)}}{B(\alpha,\beta)}
```
jgabry: @imadmali On the right-hand side the `x_i` should both be `y_i` right?

vignettes/betareg.Rmd
```diff
@@ -23,36 +23,40 @@ params:
...
+the following transformations.
 $$
 \alpha = \mu\cdot\phi \\
 \beta = (1-\mu)\cdot\phi
 $$
 
-Where $g_1(\mu) = \mathbf{X}\boldsymbol{\eta}$ where $g_1(\cdot)$ is some link function, $X$ is a $n\times K$ dimensional matrix of
-predictors, and $\eta$ is a $K$ dimensional vector of parameters associated with each predictor. 
+Where $g_1(\mu) = \mathbf{X}\boldsymbol{\eta}$ where $g_1(\cdot)$ is some link
+function, $X$ is a $n\times K$ dimensional matrix of predictors, and $\eta$ is a
```
jgabry: @imadmali `n` -> `N`

vignettes/betareg.Rmd
```diff
@@ -23,36 +23,40 @@ params:
...
 
-In the simplest case (with only one set of regressors), $\phi$ is a scalar parameter. Alternativley, it
-is possible to model $\phi$ using a second set of regressors $Z$, in which case 
-$g_2(\phi) = \mathbf{Z}\boldsymbol{\gamma}$, where $g_2(\cdot)$ is some link function and $\boldsymbol{\gamma}$ is a $L$
-dimensional vector of parameters associated with the $n\times L$ dimensional predictor matrix $Z$. 
+In the simplest case (with only one set of regressors), $\phi$ is a scalar
+parameter. Alternativley, it is possible to model $\phi$ using a second set of
+regressors $Z$, in which case $g_2(\phi) = \mathbf{Z}\boldsymbol{\gamma}$, where
+$g_2(\cdot)$ is some link function and $\boldsymbol{\gamma}$ is a $L$ 
+dimensional vector of parameters associated with the $n\times L$ dimensional
```
jgabry: @imadmali `n` -> `N`

vignettes/betareg.Rmd
```diff
@@ -61,18 +65,19 @@ dimensional vector of parameters associated with the $n\times L$ dimensional pre
...
 # Posterior
 
-When using only a *single set of regressors*, the posterior distribution of $\eta$ and $\phi$ is
-proportional to the product of the likelihood contributions, priors on the $K$ $\eta$ parameters, and $\phi$,
+When using only a *single set of regressors*, the posterior distribution of
+$\eta$ and $\phi$ is proportional to the product of the likelihood
+contributions, priors on the $K$ $\eta$ parameters, and $\phi$,
 $$
 f(\boldsymbol{\eta},\phi|\mathbf{y},\mathbf{X}) \propto 
 \prod_{i=1}^n f(y_i | \alpha,\beta) \times
```
jgabry: @imadmali `n` -> `N`

vignettes/betareg.Rmd
```diff
@@ -83,12 +88,13 @@ $$
```
jgabry: @imadmali `n` -> `N`

vignettes/betareg.Rmd
imadmali: Yes. I'll make the change, and the ones you've suggested below.

tests/testthat/test_posterior_predict.R
imadmali: I only got `posterior_predict` to work and forgot about `posterior_linpred`. I'm looking at what betareg does and it seems that `predict(betaregfit, type = "link")` (which I'm believe is what `posterior_linpred` is replicating) only returns the `mu` linear predictor even when `z` variables are declared. This doesn't really seem useful since the user can't really do anything with just `mu` if a `z` design matrix is also included in the model. Do we want both `posterior_linpred` and `linear.predictors` to return mu*phi?

```
fit <- SW(stan_betareg(yield ~ batch + temp | temp, data = GasolineYield,
                                               iter = ITER, chains = CHAINS, seed = SEED, 
                                               refresh = REFRESH))
check <- betareg::betareg(yield ~ batch + temp | temp, data = GasolineYield)

check_for_error(fit)
expect_linpred_equal(fit)

mu <- model.matrix(yield ~ batch + temp, data = GasolineYield) %*% coef(check)[1:11]
phi <- model.matrix(yield ~ temp, data = GasolineYield) %*% coef(check)[12:13]

mat <- matrix(
  cbind(
    apply(posterior_linpred(fit), 2, median),
    fit$linear.predictors,
    abs(apply(posterior_linpred(fit), 2, median)) - abs(fit$linear.predictors),
    predict(check, type = "link"),
    mu,
    phi,
    mu*phi
    ), nrow = 32
  )
colnames(mat) <- c("linpred", "linear.predictors", "diff", "betareg", "mu", "phi", "mu*phi")
mat

       linpred          linear.predictors   diff   betareg     mu          phi     mu*phi
 [1,]  0.822158294       0.738977686  0.083180607 -2.1975546 -2.1975546 4.351004  -9.5615692
 [2,]  0.588765080       0.529929302  0.058835778 -1.4723909 -1.4723909 5.370926  -7.9081032
 [3,]  0.420611907       0.320880918  0.099730989 -0.7472272 -0.7472272 6.390849  -4.7754160
 [4,]  0.298132442       0.135723778  0.162408664 -0.1049393 -0.1049393 7.294208  -0.7654494
```

bgoodri:
I think posterior_linpred() should only return alpha + X * beta or the
inverse link function thereof if the user specifies that option.

imadmali:
That makes sense for the model `y ~ x` but will it actually be used for anything if the model is `y ~ x | z`? Or are we just trying to be consistent with the betareg package?

bgoodri:
The posterior_linpred() function is not directly like any frequentist
function. It is a bit like a Bayesian version of fitted() if you apply the
inverse link function.

imadmali:
Okay, `posterior_linpred` is doing what @bgoodri suggested it should do anyway. @jgabry I think the test might have failed because iter/chains are too low.

jgabry:
Thanks. Yes it appears to just need more iterations.

jgabry:
If I'm not forgetting anything, the remaining things are:

* [ ] Implement `QR` as Ben described above
* [ ] Remove `sparse` everywhere
* [ ] Document `link.phi` default (i.e., default is `NULL` but actually `"log"`?)
* [ ] Vignette tweaks

vignettes/betareg.Rmd
imadmali: What was the issue here? `$g_1()$` is defined as "some link function" after specifying what the linear predictor is: `$g_1(\mu) = X\eta$`.

vignettes/betareg.Rmd
imadmali: Okay, I used `$\eta$` since I was using `$\alpha$` and `$\beta$` for the shape parameters, but I can use `$a$` and `$b$` instead.

vignettes/betareg.Rmd
jgabry: I think I just read  "Where g_1(mu) ..." as referring to a previous usage of g_1 above when there isn't one. It's definitely not wrong as is, but I think it would be good to write out the likelihood using the g_1(mu) and g_2(phi) terms somewhere. Maybe I'm missing it but I don't think that's written out anywhere. No big deal though.

imadmali:
> @imadmali is the default for link.phi documented somewhere? It's set to default to NULL but that's not really what it defaults to internally (seems to default to "log").

@jgabry the default is NULL when modeling phi as a scalar and `"log"` when modeling phi with a linear predictor. This was done 1) to conform to what `betareg` does and 2) to capture whether the user is specifying `z` or not. I just mentioned the default in the doc briefly. Let me know if that's sufficient.

jgabry:
@imadmali Looks good. Thanks! Just one thing: when using `expect_error` you should always put some of the expected error message so we know that the answer is for the right reason. One of your tests passes but should fail because it's a different error than the one being tested that's causing it to error.

jgabry:
@imadmali I fixed that test but there's another test that fails. I'll point you to that one in a sec

tests/testthat/test_stan_betareg.R
```diff
@@ -23,78 +23,60 @@ library(betareg)
...
+  SW(fit1 <- stan_betareg(y ~ x + z, link = "logit", seed = SEED, QR = TRUE,
                                    prior = NULL, prior_intercept = NULL,
-                                   data = dat, algorithm = "optimizing")
+                                   data = dat, algorithm = "optimizing"))
+  expect_stanreg(fit1)
+  
   fit2 <- stan_betareg(y ~ x | z, link = "logit", seed = SEED, QR = TRUE,
                                    prior = NULL, prior_intercept = NULL,
                                    data = dat, algorithm = "optimizing")
+  expect_stanreg(fit2)
```
jgabry: @imadmali this is the one that fails. I cleaned up the fake data simulation but otherwise it's the same as how you had it so either the test should fail (in which case we should test for the error) or it shouldn't fail and needs to be fixed.

jgabry:
@imadmali I think just dealing with that one test that's failing is all that's left to be done. After that we can merge this in and get `stan_betareg` released!

tests/testthat/test_stan_betareg.R
imadmali: Yeah the test should pass but I wrote it wrong. QR decomposition is not possible if the user  doesn't specify more than one x predictor or z predictor.

imadmali:
> Just one thing: when using expect_error you should always put some of the expected error message so we know that the answer is for the right reason. One of your tests passes but should fail because it's a different error than the one being tested that's causing it to error.

@jgabry I know you fixed this but do you remember where this test was? I only added QR tests and they should all print out an error message if the model isn't specified with enough predictors.

jgabry:
> @jgabry I know you fixed this but do you remember where this test was? I only added QR tests and they should all print out an error message if the model isn't specified with enough predictors.

@imadmali It was the tests for the error when x and/or z has fewer than 2 variables. You had something like this

```r
expect_error(fit <- stan_betareg(y ~ x | z, link = "logit", seed = SEED, QR = TRUE,
                                 prior = NULL, prior_intercept = NULL,
                                 data = dat, algorithm = "optimizing"))
```

but that test will pass as long as _any_ error is thrown, even if it's not the error you're testing for. We want to know that it's the QR error in particular and not some other error so that's why I changed it to something like this:

```r
  expect_error(
    stan_betareg(y ~ x | z, link = "logit", seed = SEED, QR = TRUE, data = dat),
    "'QR' can only be specified when there are multiple predictors"
  )
```
so now the test will only pass if that particular error message is thrown as opposed to any error message.

imadmali:
Ah got it thanks.

PR #153, opened 2017-01-18
Added additional association structures

sambrilleman:
Changed default priors for baseline hazard coefficients and association parameters so they are much less constraining. They now use SD of 50 for basehaz coefficients, and SD of 25 for association parameters. In the future, could potentially think about scaling them depending on the data, rather than using vague priors.

Have also added the following association structures:
(1) lagged effects
(2) cumulative effects (area under the marker trajectory)
(3) interactions between variables in dataLong and any of the other
association structure quantities (etavalue, muvalue, etaslope, etc)

Note the following for the association structures:
For (1)... For lagged values of the marker that correspond to a time earlier than baseline, the baseline marker value is assumed to apply.
For (3)... At the moment the data must be contained in the dataframe used for the
covariates in the longitudinal submodel. However I have also added an argument to stan_jm that allows the user to specify "interaction_data", which could be a data frame that contains the covariates but using a
different measurement time schedule to the data frame passed to dataLong. This would provide greater flexibility. The "interaction_data"
argument isn't properly implemented yet though.


bgoodri:
These all seem like good ideas, although I haven't looked at the code yet

bgoodri:
@sambrilleman Do you have anything more pending or can we merge the master branch as of yesterday's release into this branch?

jgabry:
Yeah I think we should keep it up to date with master as much as possible.

I agree all the changes seem reasonable as described, although I haven't
had a chance to go over the code either (it's a crazy week for us).

sambrilleman:
No, nothing pending at the moment, happy for you to merge in the up-to-date master. Cheers.

jgabry:
This has a conflict apparently, but otherwise I think we can merge it right?

Issue #173, opened 2017-03-21
cleanup betareg in continuous.stan


imadmali:
I think we no longer need the code in the generated quantities block that "catches" NaNs from the `beta_rng` function in `continuous.stan`. (Most of what I'm referring to is here https://github.com/stan-dev/rstanarm/blob/master/exec/continuous.stan#L253-L269.) The `cleanup-betareg` branch gets rid of this.

jgabry:
Thanks Imad. Are the underlying changes to Stan math library already
released in 2.14 or not until 2.15?

imadmali:
It looks like it was fixed in 2.13 https://github.com/stan-dev/math/releases/tag/v2.13.0

jgabry:
Ok great. Go ahead and make a PR whenever you think the branch is ready.

jgabry:
closed via #174

PR #174, opened 2017-03-23
cleanup-betareg

imadmali:
Cleaned up some of the code in `continuous.stan` and updated some of the tests so that they use `algorithm = "optimizing"` instead of `algorithm = "sampling"`. (Now the runtime for `devtools:tests()` is only ~5 min.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ sudo: false
 r: devel
 cache: packages
 
-r_build_args: --no-build-vignettes
-r_check_args: --no-build-vignettes
+latex: false
+r_build_args: '--no-build-vignettes'
+r_check_args: '--ignore-vignettes --no-examples'
 
 r_github_packages:
   - jimhester/covr
@@ -25,7 +26,7 @@ before_install:
 script: 
   - |
     travis_wait R CMD build .
-    travis_wait 60 R CMD check rstanarm*tar.gz
+    travis_wait 40 R CMD check rstanarm*tar.gz
 
 after_script:
   - tar -ztvf rstanarm_*.tar.gz
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -58,4 +58,4 @@ LazyData: true
 NeedsCompilation: yes
 URL: https://groups.google.com/forum/#!forum/stan-users, http://mc-stan.org/
 BugReports: https://github.com/stan-dev/rstanarm/issues
-RoxygenNote: 5.0.1
+RoxygenNote: 6.0.1
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,9 +17,10 @@ Authors@R: c(person("Jonah", "Gabry", email = "jsg2201@columbia.edu", role = "au
              person("William", "Venables", role = "cph", comment = "R/stan_polr.R"),
              person("Ben", "Goodrich", email = "benjamin.goodrich@columbia.edu", 
                     role = c("cre", "aut")))
-Description: Estimates pre-compiled regression models using the 'rstan' package, which provides
-    the R interface to the Stan C++ library for Bayesian estimation. Users specify models via the
-    customary R syntax with a formula and data.frame plus some additional arguments for priors.
+Description: Estimates pre-compiled regression models using the 'rstan' package,
+    which provides the R interface to the Stan C++ library for Bayesian estimation.
+    Users specify models via the customary R syntax with a formula and data.frame
+    plus some additional arguments for priors.
 License: GPL (>=3)
 Depends:
     R (>= 3.0.2),
@@ -57,4 +58,4 @@ LazyData: true
 NeedsCompilation: yes
 URL: https://groups.google.com/forum/#!forum/stan-users, http://mc-stan.org/
 BugReports: https://github.com/stan-dev/rstanarm/issues
-RoxygenNote: 6.0.1
+RoxygenNote: 5.0.1
--- a/exec/continuous.stan
+++ b/exec/continuous.stan
@@ -192,11 +192,8 @@ generated quantities {
     omega_int[1] = gamma_z[1] - dot_product(zbar, omega);  // adjust betareg intercept 
   }
   {
-    real nan_count; // for the beta_rng underflow issue
-    real yrep; // pick up value to test for the beta_rng underflow issue
     vector[N] eta_z;
     #include "make_eta.stan" // defines eta
-    nan_count = 0;
     if (t > 0) {
       #include "eta_add_Zb.stan"
     }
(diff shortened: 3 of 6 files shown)


imadmali:
closes #173

jgabry:
@imadmali Looks good. Do all the unit tests for the functions using continuous.stan (e.g. stan_glm) still pass too?

imadmali:
Based solely on the changes made to continuous.stan all the tests pass. But for some reason, when I use `algorithm = "optimizing"` instead of `algorithm = "sampling"` when modeling `phi` with the identity function there's a negative shape parameter error. Then the corresponding tests [here](https://github.com/stan-dev/rstanarm/blob/cleanup-betareg/tests/testthat/test_stan_betareg.R#L135-L153) fail. I'm not sure why this is happening. Examples below...

This runs:
```
N <- 200
dat <- data.frame(x = rnorm(N, 2, 1), z = rnorm(N, 2, 1))
mu <- binomial(link = "logit")$linkinv(1 + 0.2*dat$x)
phi <- dat$z - min(dat$z) + 8
dat$y <- rbeta(N, mu * phi, (1 - mu) * phi)
fit <- stan_betareg(y ~ x | z, link = "logit", link.phi = "identity",
                    prior = NULL, prior_intercept = NULL,
                    prior_z = NULL, prior_intercept_z = NULL,
                    data = dat, algorithm = "sampling", chains = 2, iter = 100)
```

But this doesn't run:
``` 
N <- 200
dat <- data.frame(x = rnorm(N, 2, 1), z = rnorm(N, 2, 1))
mu <- binomial(link = "logit")$linkinv(1 + 0.2*dat$x)
phi <- dat$z - min(dat$z) + 8
dat$y <- rbeta(N, mu * phi, (1 - mu) * phi)
fit <- stan_betareg(y ~ x | z, link = "logit", link.phi = "identity",
                    prior = NULL, prior_intercept = NULL,
                    prior_z = NULL, prior_intercept_z = NULL,
                    data = dat, algorithm = "optimizing")
```

jgabry:
Hmm, what happens if you initialize the shape parameter to something
reasonable?

imadmali:
I'm assuming you were referring to `phi` and yeah the following works:
```
fit <- stan_betareg(y ~ x | z, link = "logit", link.phi = "identity",
                    prior = NULL, prior_intercept = NULL,
                    prior_z = NULL, prior_intercept_z = NULL,
                    data = dat, algorithm = "optimizing",
                    seed = SEED)
```

jgabry:
What's the difference between that and what you had before? That looks like
the same thing as in the previous comment or maybe I'm not noticing
something.

I was just saying that it might be initializing poorly if left to do random
inits and then optimization could fail. Could fail for other reasons too of
course.

imadmali:
> What's the difference between that and what you had before? That looks like
the same thing as in the previous comment or maybe I'm not noticing
something.

Yeah it's similar but now I've fixed `phi`. Since the shape parameters enter `beta_lpdf` in a complicated way it sometimes helps to treat `phi` as a scalar (from what I remember) to generate reasonable data. I probably should play around with more edge cases since these models are really sensitive to the data.

> I was just saying that it might be initializing poorly if left to do random
inits and then optimization could fail. Could fail for other reasons too of
course.

Yeah that could've been it. Regardless, the model specification is a poor from an applied statistics point of view. I remember talking with @bgoodri about this and typically what makes sense is to only model `mu` as the linear predictor and fix `phi` as a scalar.

jgabry:
Ok we should probably include a warning if someone tries to fit a model
that we know doesn't make sense.

Actually we probably just shouldn't allow it.

imadmali:
As in we shouldn't allow users to fit a model with `link.phi = "identity"`?

Also, the following tests fail on the cleanup-betareg branch. I also noticed that the same tests fail on the current master branch. Are they failing for you, or am I doing something wrong when I switch branches?

```
Failed ------------------------------------------------------------------------------------------------------------------------------
1. Failure: print and summary methods ok for mcmc and vb (@test_methods.R#597) ------------------------------------------------------
print(s) does not match "Posterior sample size: 500".
Actual value: "\nModel Info:\n\n function:  stan_glmer\n family:    binomial [logit]\n formula:   cbind(incidence, size - incidence) ~ size + period + (1 | herd)\n algorithm: sampling\n priors:    see help('prior_summary')\n sample:    500 (posterior sample size)\n num obs:   56\n groups:    herd (15)\n\nEstimates:\n                                      mean   sd   2.5%   25%   50%   75%   97.5%\nsize                                 0.0    0.0  0.0    0.0   0.0   0.0   0.1   \nperiod2                             -1.0    0.3 -1.7   -1.2  -0.9  -0.7  -0.3   \nperiod3                             -1.1    0.3 -1.7   -1.3  -1.1  -0.9  -0.4   \nperiod4                             -1.5    0.5 -2.6   -1.8  -1.5  -1.2  -0.6   \nb[(Intercept) herd:1]                0.7    0.5 -0.2    0.3   0.7   0.9   1.6   \nb[(Intercept) herd:2]               -0.4    0.5 -1.3   -0.7  -0.3  -0.1   0.4   \nb[(Intercept) herd:3]                0.4    0.4 -0.4    0.2   0.4   0.7   1.1   \nb[(Intercept) herd:4]     ... <truncated>


2. Failure: recommend_QR_for_vb produces the right message (@test_misc.R#495) -------------------------------------------------------
recommend_QR_for_vb() produced no output


3. Failure: pp_validate runs for very quick example (@test_pp_validate.R#36) --------------------------------------------------------
gg <- pp_validate(fit, nreps = 2, seed = SEED) produced no output


4. Failure: predict ok for binomial (@test_predict.R#77) ----------------------------------------------------------------------------
pg$fit not equal to ps$fit.
41/51 mismatches (average diff: 0.158)
[1] -2.82 - -3.04 == 0.226
[2] -2.69 - -2.91 == 0.215
[3] -2.57 - -2.77 == 0.205
[4] -2.44 - -2.64 == 0.195
[5] -2.31 - -2.50 == 0.185
[6] -2.19 - -2.36 == 0.175
[7] -2.06 - -2.23 == 0.164
[8] -1.94 - -2.09 == 0.154
[9] -1.81 - -1.96 == 0.144
...


DONE ================================================================================================================================
```

jgabry:
The errors from the tests for the print and summary methods are fine to
ignore. Those are my fault. I'll take care of them.

With the phi thing, I just meant that if we know optimization won't work
for a particular model we shouldn't allow it. But If we can get it to work
then great.

jgabry:
Ok I fixed all of the failing tests you found except for the last one (predict for binomial). I don't know why that would be failing. @bgoodri any ideas?

bgoodri:
No, but I noticed some unrelated failures when working on various branches
recently.

imadmali:
Sorry, but I only got around to looking at this now and it looks like the latest commit that got merged was the one that used `algorithm = "optimizing"` in the tests for the `link.phi = "identity"` model. So this test will (probably) always fail: https://github.com/stan-dev/rstanarm/blob/master/tests/testthat/test_stan_betareg.R#L135-L153.

Also, @jgabry mentioned stopping users from being able to fit a model with `link.phi="identity"`, which I implemented in the latest commit of the cleanup-betareg branch.

Also, I updated the `link.phi = "sqrt"` tests to use `algorithm = "optimizing"`. The only errors that seem to come up are sampling proposal rejections (see below), which I don't think are picked up by testthat so that should be fine.

```
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception thrown at line -1: beta_log: Second shape parameter[6] is 0, but must be > 0!
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Rejecting initial value:
  Error evaluating the log probability at the initial value.

Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception thrown at line -1: beta_log: Second shape parameter[1] is 0, but must be > 0!
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Rejecting initial value:
  Error evaluating the log probability at the initial value.

...
```

PR #179, opened 2017-04-10
Vignette about prior specification

### LINKED ISSUE #167: Add vignette about priors
We should have a vignette that covers all the basics of how priors work in rstanarm and what our general recommendations are. Currently this information is spread out across various documentation pages in the package and there are bits and pieces in the other vignettes. 

@bgoodri I have a rough draft of this on a branch from a while ago (still needs some work) that I will push to GitHub once this potential conflict between our licenses and the ToS is sorted out.

jgabry:
Closes #167

There's a lot more to add to this vignette but I think there's probably enough in it already to warrant putting it in a release. We can continue to refine it in future releases. But it could use a second pair of eyes first.

@bgoodri Any additions or corrections?

### DIFF
--- a/R/prior_summary.R
+++ b/R/prior_summary.R
@@ -256,7 +256,7 @@ used.sparse <- function(x) {
       if (!is.null(p$df))
         p$df <- .format_pars(p$df, .f1)
       if (!is.null(p$adjusted_scale))
-        p$adjusted_scale <- .format_pars(p$adjusted_scale, .f1)
+        p$adjusted_scale <- .format_pars(p$adjusted_scale, .f2)
     } else if (p$dist %in% c("hs_plus")) {
       p$df1 <- .format_pars(p$df, .f1)
       p$df2 <- .format_pars(p$scale, .f1)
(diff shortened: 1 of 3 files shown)


imadmali:
This is really helpful @jgabry.

It's not a big deal but somewhere I would mention that the user can specify a vector of parameters in certain prior distributions (e.g. `prior = normal(c(0,1),c(2,2)`). This wasn't obvious to me when I first started using rstanarm. It's implied in the **Informative Priors Section** but it might be nice to explicitly state it somewhere.

It would also be nice to have a table that outlines the **Details** section of `?priors`. Something like,
```
Family | Family Member | Models
-----------------------------------
R2       | R2()        | stan_lm, stan_aov, stan_polr
-----------------------------------
Student t| normal()    | stan_glm, stan_glmer, stan_betareg
...
```
Again, none of this particularly urgent, but it's stuff that I would've appreciated having when I first started using rstanarm.

jgabry:
@imadmali Thanks. I definitely plan on adding more examples of using vector
arguments with the prior functions.

Also, feel free to add to it if there are other things you think should be
in there.

imadmali:
Will do.

PR #180, opened 2017-04-11
Feature/reorganize tests

### LINKED ISSUE #175: Reorganize unit tests
#### Summary:

The current unit tests are too duplicative and take too much time to run. We should reorganize most of them so that a file estimates a particular model and then tests `posterior_predict`, `loo`, etc.

#### Description:

Take test_loo.R for example. It has a "test that loo/waic for stan_glm works", which estimates six Stan models that are already being estimated elsewhere in the test suite. We should have

- test_gaussian.R
- test_binomial.R
- test_poisson.R
- test_negbin.R
- test_gamma.R
- test_inverse_gaussian.R

etc. that each obtain a reasonable number of draws from a posterior distribution once and then test that all the post-estimation things work.

#### Reproducible Steps:

Look at the tests

#### RStanARM Version:

2.14.1

#### R Version:

3.3.3

#### Operating System:

Debian

bgoodri:
closes #175


bgoodri:
Things are better now but still need more changes

PR #196, opened 2017-07-01
Feature/clogit

bgoodri:
Not quite ready yet


jgabry:
Cool model. Going through the diff I don't see anything obviously problematic but I haven't had a chance to actually try it yet.

bgoodri:
You can also do
```
nascar <- read.table("http://ww2.amstat.org/publications/jse/datasets/
nascard.dat.txt",
                     stringsAsFactors = FALSE)
colnames(nascar) <- c("RaceID", "Year", "RaceOfYear",
                      "Finish", "Start", "Laps", "Money",
                      "Cars", "Make", "Name")
post <- stan_clogit(Finish == 1 ~ Start + Start:Cars + (1 | Name), data =
nascar, strata = RaceID, QR = TRUE)
```

R/posterior_linpred.R
```diff
@@ -43,6 +43,14 @@
 #'   transformed) linear predictor. The exception is if the argument \code{XZ}
 #'   is set to \code{TRUE} (see the \code{XZ} argument description above).
 #'   
+#' @note For models estimated with \code{\link{stan_clogit}}, the number of 
+#'   successes per stratum is ostensibly fixed by the research design. Thus,
+#'   when calling \code{posterior_linpred} with new data and \code{transform =
+#'   TRUE}, the \code{data.frame} passed to the \code{newdata} argument must
+#'   contain an outcome variable and a stratifying factor, both with the same
+#'   name as in the original \code{data.frame}. Then, the probabilities will
```
jgabry: Currently the `strata` argument to `stan_clogit` can either be the (unquoted) name of a variable (which is then looked for in `data`) or an actual factor that doesn't have to correspond to a column in `data`. In the second case there is no variable name in the data frame for the strata variable so how will `newdata` work?

R/posterior_predict.R
```diff
@@ -82,6 +82,12 @@
 #'   probably with \code{successes} set to \code{0} and \code{trials} specifying
 #'   the number of trials. See the Examples section below and the
 #'   \emph{How to Use the rstanarm Package} for examples.
+#' @note For models estimated with \code{\link{stan_clogit}}, the number of 
+#'   successes per stratum is ostensibly fixed by the research design. Thus, when
+#'   doing posterior prediction with new data, the \code{data.frame} passed to
+#'   the \code{newdata} argument must contain an outcome variable and a stratifying
+#'   factor, both with the same name as in the original \code{data.frame}. Then, the
```
jgabry: See comment on `posterior_linpred` above.

jgabry: For something like this I think we should have an example of the correct way to do it like we do for the binomial case

tests/testthat/test_stan_clogit.R
```diff
@@ -0,0 +1,69 @@
...
+CORES <- 1
+REFRESH <- 0
+
+threshold <- 0.01
+
+source(file.path("helpers", "expect_stanreg.R"))
+
+context("stan_clogit")
+
+fit <- stan_clogit(case ~ spontaneous + induced, strata = stratum, prior = NULL,
```
jgabry: Probably should test the multilevel syntax also. If we don't have a comparison for that, we can at least test that it runs.

R/stan_clogit.R
```diff
@@ -0,0 +1,174 @@
...
+#'   the variable passed to the \code{strata} argument.
+#'   
+#'   The \code{formula} may have group-specific terms like in 
+#'   \code{\link{stan_glmer}} but should not allow the intercept to vary by the
+#'   stratifying variable, since there is no information in the data with which
+#'   to estimate such deviations in the intercept.
+#'   
+#' @seealso The vignette for Bernoulli and binomial models.
+#' 
+#' @examples
```
jgabry: If lme4 syntax is allowed I think there should be an example

inst/chunks/bernoulli_likelihoods.stan
```diff
@@ -72,3 +72,56 @@
...
+
+  /** 
+   * Log-normalizing constant in the clogit case
+   *
+   * @param N_j Integer number of observations in the j-th group
+   * @param D_j Integer number of successes in the j-th group
+   * @param eta_j Vector of linear predictions in the j-th group
+   * @return A scalar that normalizes the probabilities on the log-scale
+   */
+  real log_clogit_denom(int N_j, int D_j, vector eta_j);
```
jgabry: Looks like a duplicate line that was maybe originally planning on being used for external c++ code. I think it can be removed though, right?

jgabry:
Also anything special about how these models should behave with `pp_check`?

R/stan_clogit.R
jgabry: Maybe the nascar example you put in the comments above?

jgabry:
@bgoodri Now that stan_nlmer is merged I fixed the merged conflicts on the stan_clogit branch. So I think stan_clogit is ready to go as soon as that issue with newdata and strata is sorted out (see comments above).

R/posterior_linpred.R
bgoodri: I am going to prevent the case where `stratum` involves something outside the `data.frame`.

R/stan_clogit.R
bgoodri: I am adding `(1 | education)` to the example.

inst/chunks/bernoulli_likelihoods.stan
bgoodri: The `log_clogit_denom` function is recursive.

R/posterior_predict.R
bgoodri: It is sort of artificial but I am putting it into the example for `stan_clogit`.

Issue #200, opened 2017-07-07
Regularized horseshoe


jpiironen:
#### Summary:
Wish: implement the regularized horseshoe prior.

#### Description:
Would be great to have the regularized horseshoe implemented in prior `hs`. The paper is now in arXiv (https://arxiv.org/abs/1707.01694) and the new prior is discussed in Section 2.3. This should be easy to add on top of the current implementation of the `hs`-prior, and there's an example code in Appendix C of the paper.

I think using the Student-t slab is fine at least for the time being, and the syntax could be something like 
`prior <- hs(df=1, global_df=1, global_scale=0.01, slab_df=4, slab_scale=2)`
I also think `slab_df = 4` would be a good default value for the degrees of freedom but for `slab_scale` you could probably use the same default value as you use in `student_t`-prior (which seems to be 2.5 except for probit regression if I understood right). I don't have much preferences regarding this.

Also, although it breaks the backward compatibility, I would now change the default shrinkage degrees of freedom to `df = 1` (horseshoe) as with the regularized horseshoe there is no more reason to recommend using `df > 1` (see Sec. 2.4 of the paper).

Thanks a lot!

bgoodri:
What should become of hs_plus ()?

bgoodri:
Also, I am not understanding why global_scale is still an argument to the
function rather than p_0.

jpiironen:
`hs_plus` could be regularized as well. 

For the horseshoe, the regularization is achieved as follows
```
lambda_tilde = sqrt( c^2 * square(lambda) ./ (c^2 + tau^2*square(lambda)) )
beta ~ N(0, lambda_tilde * tau)
lambda ~ C+(0,1)
```
and for the horseshoe+ like this
```
lambda_tilde = sqrt( c^2 * square(lambda .* eta) ./ (c^2 + tau^2*square(lambda .* eta)) )
beta ~ N(0, lambda_tilde * tau)
lambda ~ C+(0,1)
eta ~ C+(0,1)
```

[...]

PR #205, opened 2017-07-17
hs(): Switch to regularized horseshoe

bgoodri:
I think this is correct, but we have to decide what to do about `hs_plus()` and what should be the arguments to `hs()`.


jpiironen:
In case you're still thinking about what to do about `hs_plus`, I posted my idea in the issue thread #200 (as you probably already noticed).

avehtari:
The paper with regularized horseshoe has been accepted to Electronic Journal of Statistics. I'm wondering what is the status of this PR?

jgabry:
I think all the open PRs (or at least almost all of them, including this
one) will get merged before the release. But we need to get rstan out
before we do rstanarm.

bgoodri:
@jpiironen @avehtari This may be ready to merge but it is sort of hard to test that it is correct due to all the reparameterizing. So, it would be good if you could look over the diff.

jpiironen:
I went through the changes and didn't spot any mistakes, so I think this is good to go if you also think so!

PR #209, opened 2017-07-25
Drop warmup iterations from returned stanfit object

sambrilleman:
I had previously been retaining the warmup iterations (since they were sometimes useful for diagnostic and debugging reasons). But it is probably preferable to drop them so that the size of the returned object is smaller.

You may also be interested in some of the other defaults in the `set_sampling_args_for_jm` function, since they differ somewhat from the `set_sampling_args` function. Mostly because the computation time for the joint model is so long, and therefore 4 chains at 2000 iterations and high treedepth would be infeasible unless in parallel.


bgoodri:
I think treedepth, adapt_delta, etc. should have defaults that are
appropriate for the particular model. At least that is what we have have
done with adapt_delta for the other models in rstanarm. I am a bit more
concerned about changing the default number of iterations and chains.

sambrilleman:
Ok, seems fair enough. [This](https://github.com/stan-dev/rstanarm/pull/209/commits/98bcdac40436f1922df9be278f273073132f16dc) commit returns chains, iterations, and refresh all to the rstan defaults.

bgoodri:
I am still trying to go through all of this. The unit tests are really comprehensive but they almost double the testing time which CRAN won't like, so I had to go ahead and disable test_stan_jm.R for CRAN. I think the branch also needs a vignette. There are a bunch of other ones that you can Save As jm.Rmd or whatever and then replace the text / code.

sambrilleman:
Thanks, Ben. No worries about disabling the tests -- I guess we can still run them on locally and perhaps on Travis?

With regard to the vignette, I started trying to pull something together this afternoon, but haven't got that far before I have run out of time for the day. And heading away on leave tomorrow. See [this](https://github.com/sambrilleman/rstanarm/commit/8063ab72fb0fa3990775c9a79dba33f7702e6754) commit -- it contains what I have done so far, so happy to let you decide whether it would be worth including as-is, or whether we should wait till I have finished a bit more of it (which I won't get a chance to do for a little while, not sure exactly when!)

bgoodri:
This was subsumed by merging #237

PR #211, opened 2017-07-26
bayes_R2 function

jgabry:
@bgoodri Look ok to you? Right now this is just the function, I didn't add anything to the printed output after fitting a model. But let's get the function done and then we can eventually decide if/when/how we want to display it. Also whenever we put the little paper on arXiv I'll add a link to it from the function documentation.


paul-buerkner:
Maybe you could make a generic for it in `rstantools`, so that other packages can use the same function name.

jgabry:
Thanks that slipped my mind. Opened an issue to add the generic: https://github.com/stan-dev/rstantools/issues/8

jgabry:
@bgoodri To get this to run in rstanarm you have to first install **rstantools** from GitHub so that the new generic `bayes_R2` is available.

jgabry:
__rstantools__ with `bayes_R2` generic is now on CRAN. @bgoodri this can now be merged if it looks ok to you.

PR #216, opened 2017-08-05
Print number of predictors for most single-level models

jgabry:
For models fit with `stan_glm` and `stan_lm` only, this PR adds a line to the `print.stanreg` and `summary.stanreg` output giving the number of predictors. I've used `length(coef(object))` to count the predictors, which will count the intercept as a predictor. In the example below the number of predictors is computed to be 4: the intercept, the `wt` variable, and the two levels of `cyl` (third level is dropped because there's an intercept). 

```r
fit <- stan_glm(mpg ~ wt + factor(cyl), data = mtcars)
print(fit)
```
```
stan_glm
 family:       gaussian [identity]
 formula:      mpg ~ wt + factor(cyl)
 observations: 32
 predictors:   4
------
(rest of print output the same as before)
```

@bgoodri If Andrew prefers to use `n` and `k` instead of `observations` and `predictors` we can change it later I guess, but I think I prefer using `observations` and `predictors`. Do you have a preference?

### DIFF
--- a/R/print-and-summary.R
+++ b/R/print-and-summary.R
@@ -69,7 +69,7 @@ print.stanreg <- function(x, digits = 1, ...) {
   cat("\n family:      ", family_plus_link(x))
   cat("\n formula:     ", formula_string(formula(x)))
   cat("\n observations:", nobs(x))
-  if (x$stan_function %in% c("stan_glm", "stan_glm.nb", "stan_lm"))
+  if (isTRUE(x$stan_function %in% c("stan_glm", "stan_glm.nb", "stan_lm")))
     cat("\n predictors:  ", length(coef(x)))
   
   cat("\n------\n")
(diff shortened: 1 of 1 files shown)


bgoodri:
Observations definitely rather than n or N. Andrew has some overly strong
preference about the word "predictors". I think to him "state ID" would be
one predictor but implies 50 covariates or something like that.

jgabry:
Hmm ok. I don't really want to use "k" but I guess we could use
"covariates" instead of "predictors". I'll leave it as "predictors" for now
and we can ask him on Monday what he prefers and then change and merge
this.

jgabry:
@bgoodri Just talked to Andrew about the terminology and he's ok with using number of "predictors".

bgoodri:
OK

PR #221, opened 2017-08-14
Use consistent family numbers across Stan files

### LINKED ISSUE #218: Improve consistency across .stan files
The .stan files in rstanarm will never be super easy to read, but they're currently harder to follow than they need to be. One relatively simple improvement would be to not use the same names/values for things in different Stan files when the interpretation is different. 

For example in the chunk `data_glm.stan` we have

```stan
...
  // family (interpretation varies by .stan file)
  int<lower=1> family;
...
```
in order to allow things like `family=1` referring to Gaussian in `continuous.stan` and Poisson in `count.stan`, etc.  In hindsight, I think doing it this way was a mistake. There's no reason `family` needs to start at `1` in each file. So we could have `family=1` always be Gaussian and if there's no Gaussian family in a certain file then there will be no `family=1`.

This would also cut down on R code (some of `stan_glm.fit` could be simplified) and Stan code (more Stan code could be shared across files, e.g. stuff from generated quantities).  @imadmali has also run into some examples of this when working on the spatial models, where he's had to write code that could have been shared.

jgabry:
This PR resolves #218. The following integer family codes are now used throughout the Stan files in  rstanarm instead of the interpretation varying by file:

*   1 = gaussian
*   2 = gamma
*   3 = inverse gaussian
*   4 = beta
*   5 = binomial
*   6 = poisson
*   7 = negative binomial
*   8 = poisson w/ gamma noise (not currently exposed but exists in count.stan)

This should make existing Stan files more clear (not necessarily individually but as a collection) and it will also mean that @imadmali can use more existing code in his spatial models, which use most of the families. 

I think we should ultimately do this with link functions and anything else that has a varying interpretation across files, but this is a start.


jgabry:
This passes all tests when I run them locally but I had to fix the `stan_nlmer` tests. It had `prior=NULL` but was missing `QR=TRUE` and there were hundreds of divergences and nonsense results. I don't think that's related to this PR, but the test probably hasn't ever passed on the master branch either. We just didn't know because travis doesn't work.

PR #258, opened 2018-02-28
Remove startup ggplot2 theme set and add theme to plot functions

### LINKED ISSUE #257: rstanarm changes ggplot2 theme at start-up
Hi folks,

Firstly, thanks for the great package and all the hard work that's gone into it. It's really a treat to use.

`rstanarm` changes the `ggplot2` theme to `bayesplot::theme_default()` at start-up. I believe this is misguided, for two reasons:
1. It globally changes the `ggplot2` theme, even though I can say as an analyst, plots related to my model are only a subset of the plots that I am making. Obviously, this can just as easily be overriden by adding the `ggplot2` default theme to those plots, but I don't think the default of this package should  "break" the theme of other plots.
1. It's not really in the spirit of `ggplot2`'s layer-by-layer API; if you want the `bayesplot` theme, you should just add it, particularly since it's only a line of code, and otherwise leave it up to the user.

I do appreciate that `bayesplot` has a nice aesthetic, and that it's also nice to have a distinct look based on that for `rstanarm` to have, but I think this is the wrong way to go about it. I'm submitting a PR that removes the startup code and adds a quick function to set it. It also directly adds `bayesplot::theme_default()` to `plot.stanreg` and `pairs.stanreg`.

Thanks!
Malcolm

malcolmbarrett:
Fixes #257

### DIFF
--- a/R/plots.R
+++ b/R/plots.R
@@ -474,3 +474,14 @@ pairs.stanreg <-
   }
   return(max_td)
 }
+
+
+#' Quickly set the default ggplot2 theme to the bayesplot style
+#'
+#' This is a convenience function to quickly change the default \code{ggplot2}
+#' theme to the default theme from the \code{bayesplot} package.
+#'
+#' @export
+set_bayes_theme <- function() {
+  ggplot2::theme_set(bayesplot::theme_default())
+}
--- a/R/plots.R
+++ b/R/plots.R
@@ -449,6 +449,10 @@ pairs.stanreg <-
     }
     posterior <- round(posterior, digits = 12)
     
+    gg_theme <- ggplot2::theme_get()
+    ggplot2::theme_set(bayesplot::theme_default())
+    on.exit(ggplot2::theme_set(gg_theme))
+    
     bayesplot::mcmc_pairs(
       x = posterior, 
       np = bayesplot::nuts_params(x),  
@@ -456,7 +460,7 @@ pairs.stanreg <-
       max_treedepth = .max_treedepth(x),
       condition = condition,
       ...
-    ) + bayesplot::theme_default()
+    )
     
   }
 
--- a/R/plots.R
+++ b/R/plots.R
@@ -168,7 +168,8 @@ plot.stanreg <- function(x, plotfun = "intervals", pars = NULL,
   
   fun <- set_plotting_fun(plotfun)
   args <- set_plotting_args(x, pars, regex_pars, ..., plotfun = plotfun)
-  do.call(fun, args)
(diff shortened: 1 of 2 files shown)


jgabry:
@malcolmbarrett Thanks for this! I haven't looked through it in detail yet, but I will. This issue with bayesplot setting the theme dates back to this closed issue at the bayesplot repo: 

https://github.com/stan-dev/bayesplot/issues/87

But I've gotten a lot of complaints about it recently and it bothers me too. So I'm regretting that decision. I understand why it was suggested that I make the change in the first place, but I think it probably wasn't the right call. Anyway, I'm now thinking it might be worthwhile to have a `bayesplot_theme_set` function that does what `ggplot2::theme_set` does but only affects the theme for bayesplot plots. But I haven't tried to implement that yet.  What do you think?

malcolmbarrett:
Ah, I do see what the idea behind it was. I hadn't realized that `theme_set()` can't affect a plot that has a theme hard-coded, although I wonder if ultimately that is an issue that should be brought up in `ggplot2` itself, some sort of function that works within namespaces but prioritizes the global setting.

This PR included a quick function to set it universally, but it's interesting to think about how it could be handled locally in `bayesplot`.

Something like this? (Basically a small modification of how `ggplot2` handles it.)
``` {r}
bayes_theme_env <- new.env(parent = emptyenv())
bayes_theme_env$current <- bayesplot::theme_default()

example_hist <- function(...) {
  ggplot2::qplot(...) + bayesplot_theme_get()
}

bayesplot_theme_set <- function(new) {
  missing <- setdiff(names(ggplot2::theme_gray()), names(new))
  if (length(missing) > 0) {
    warning("New theme missing the following elements: ",
            paste(missing, collapse = ", "), call. = FALSE)
  }

  old <- bayes_theme_env$current
  bayes_theme_env$current <- new
  invisible(old)
}

bayesplot_theme_get <- function(variables) {
  bayes_theme_env$current
}
```

`example_hist(1:5)` uses the `bayesplot` theme, but can be overridden with `bayesplot_theme_set()`. It also doesn't require a call on package startup. It would be easy, as well, to add a check to see if the user has used `set_theme()` to change the default to anything that is different than `theme_gray()` and use that instead of `bayesplot_theme_default()`.

I also realize now that this is really a `bayesplot` issue that has seeped into `rstanarm`, if you would prefer to move the conversation over there.

tjmahr:
Thanks for sharing the demo code. You saved us all some time for having to study how ggplot2 does it.

jgabry:
Thanks!!

PR #259, opened 2018-03-01
Update glmer.Rmd

ifengji:
fixed a typo


jgabry:
Thanks for fixing that!

PR #263, opened 2018-03-20
Fix serious bug with etaauc & muauc association structures

sambrilleman:
The etaauc and muauc association structures had a bug which meant that the baseline hazard estimate was wrong for joint models estimated with assoc = "etaauc" or assoc = "muauc". The regression coefficients were still correct, but the error in the baseline hazard meant that survival predictions were wrong. These errors can be seen, for example, by fitting a joint model with assoc = "etaauc" and then checking posterior predictions via the ps_check() function.

The problem was that the intercept term wasn't included in the linear predictor for the longitudinal submodel when calculating the AUC for the longitudinal submodel, leading to a shift in the linear predictor for the event submodel that wasn't corrected post-estimation.

It was a simple copy and paste error, that led to a serious bug that I hadn't picked up on. Resolved as per this commit.


jgabry:
Thanks Sam. We’re gonna do a release very soon so we’ll get this merged before we do that.

PR #264, opened 2018-03-22
mrp in rstanarm vignette

lauken13:
-takes 1-2 hours to compile/run(on an older laptop)
-undecided whether interaction in model should be varying or constant
-separate bibliography/references file.

### DIFF
--- a/vignettes/bibliography.bib
+++ b/vignettes/bibliography.bib
@@ -1,0 +1,112 @@
+  @Misc{RStanArm,
+    title = {rstanarm: {Bayesian} applied regression modeling via
+      {Stan}.},
+    author = {{Stan Development Team}},
+    note = {R package version 2.13.1},
+    year = {2016},
+    url = {http://mc-stan.org/},
+  }
+  
+  @article{little1993post,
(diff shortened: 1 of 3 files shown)


jgabry:
Talked to @lauken13 and there are still a few things to change in this so we're not going to include this with this release but wait for the next one.

lauken13:
New vignette takes 15minutes to compile and incorporates feedback from @jgabry. Let me know if there's anything else that you'd like changed.

saudiwin:
what a great vignette!

jgabry:
@saudiwin Glad you like it! 

@lauken13 Speaking of this vignette, I found a few things that I had missed that we need to change in order to get this vignette on CRAN. Shouldn't take long at all, but we should chat about it briefly when we next meet so we can get this released.

jgabry:
* [ ] Change inline R code to chunks

* [ ] Add towards the top where similar things are:  {r, child="children/SETTINGS-knitr.txt"}

lauken13:
Updated based on suggestions, also changed authorship

PR #278, opened 2018-05-02
Remove theme_set() on startup and change message

### LINKED ISSUE #257: rstanarm changes ggplot2 theme at start-up

malcolmbarrett:
Closes #257

Reflects changes to `bayesplot` approach to themes in stan-dev/bayesplot#149

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -31,7 +31,7 @@
   packageStartupMessage("Thus, R scripts should specify priors explicitly, even if they are just the defaults.")
   packageStartupMessage("- For execution on a local, multicore CPU with excess RAM we recommend calling")
   packageStartupMessage("options(mc.cores = parallel::detectCores())")
-  packageStartupMessage("- Plotting theme set to bayesplot::theme_default().")
-  ggplot2::theme_set(bayesplot::theme_default())
+  packageStartupMessage("- bayesplot theme set to bayesplot::theme_default(). Change it")
+  packageStartupMessage("  with ggplot2::theme_set() or bayesplot::bayesplot_theme_set()")
 }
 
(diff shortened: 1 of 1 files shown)


jgabry:
Thanks for this! I’ll review the bayesplot one first and once that’s merged I’ll review/merge this one for rstanarm.

jgabry:
And don’t worry about Travis-ci errors unless they occur before it times out. But it will probably time out and error even if all is ok, so we’ll have to check it locally.

malcolmbarrett:
I'm going to update this to match the bayesplot startup message. This one is much simpler, but anything else that needs to be done before merging?

malcolmbarrett:
Although, now that I say it, it makes less sense for the vignettes that have raw ggplot2 to be using theme_default() without explicitly calling it in the code since rstanarm will no longer change the theme for all ggplots.

bgoodri:
@jgabry What is up with this PR?

PR #279, opened 2018-05-09
Fix typos in vignettes.

mcol:
This fixes a number of typos in the source files for the vignettes. In one or two cases, it also fixes broken links between the vignettes.


jgabry:
@mcol thanks for fixing all these!

PR #287, opened 2018-05-14
Print aux separately

### LINKED ISSUE #284: Add space before printing sigma and aux, so the don't look like coef
#### Summary:
When doing `print(fit)` add space before printing sigma and aux, so the don't look like coef

The current output
```
            Median MAD_SD
(Intercept) -1.1    1.2  
height       0.0    0.0  
male         0.9    0.2  
sigma        1.9    0.0  
```
The proposed output
```
            Median MAD_SD
(Intercept) -1.1    1.2  
height       0.0    0.0  
male         0.9    0.2  
 sigma       1.9    0.0  
```
Other suggestions how to separate sigma/aux from coefs are also welcome.

jgabry:
Separates the printing of auxiliary variables (and marks them as such) in order to distinguish them from regression coefficients in the printed output. 

```
stan_glm
 family:       gaussian [identity]
 formula:      mpg ~ wt + cyl
 observations: 32
 predictors:   3
------
            Median MAD_SD
(Intercept) 39.7    1.7  
wt          -3.2    0.7  
cyl         -1.5    0.4  

Auxiliary parameter(s):
      Median MAD_SD
sigma 2.6    0.4   
```

This should be clear to the user since sigma (and the other auxiliary parameters for other GLMs) are specified via `prior_aux` and are referred to as auxiliary parameters in the doc. 

Closes #284

### DIFF
--- a/R/print-and-summary.R
+++ b/R/print-and-summary.R
@@ -72,19 +72,28 @@ print.stanreg <- function(x, digits = 1, ...) {
   cat("\n family:      ", family_plus_link(x))
   cat("\n formula:     ", formula_string(formula(x)))
   cat("\n observations:", nobs(x))
-  if (isTRUE(x$stan_function %in% c("stan_glm", "stan_glm.nb", "stan_lm")))
+  if (isTRUE(x$stan_function %in% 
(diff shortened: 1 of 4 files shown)


jgabry:
The PR making mean_PPD optional will require this so merging this PR first

PR #292, opened 2018-05-17
Function "nsamples()" for stanreg models

strengejacke:
This is a short and convenient wrappter to extract the number of samples, similar to what @paul-buerkner has implemented in _brms_ (https://github.com/paul-buerkner/brms/blob/master/R/generics.R).

The generic and the _brmsfit_-methods are both in the package _brms_, but maybe it would make sense to move the generic to _rstantools_?

### DIFF
--- a/R/nsamples
+++ b/R/nsamples
@@ -1,0 +1,10 @@
+#' @importFrom purrr map_dbl
+#' @export
+nsamples.stanreg <- function(x, incl_warmup = FALSE, ...) {
+  if (incl_warmup)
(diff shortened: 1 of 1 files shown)


jgabry:
Thanks @strengejacke. @paul-buerkner do you want to make a PR to rstantools to move the generic there? Then we can add this to rstanarm too.

paul-buerkner:
Will do tomorrow.

bgoodri:
Does this need anything else before merging?

PR #299, opened 2018-06-10
Parameterize Joint Models vignette text

statwonk:
I was reading through @sambrilleman's awesome work on Joint Models in `rstanarm` (https://cran.r-project.org/web/packages/rstanarm/vignettes/jm.html) when I noticed a small discrepancy between the vignette text and model output. This change parameterizes the text as a function of the model object to keep the two in sync.


bgoodri:
We usually cannot use `r foo()` constructs in the text of an RMarkdown file because the code to which they refer does not get executed on CRAN (because it takes too long to run).

statwonk:
Ahh, makes sense @bgoodri. I replaced with static version of the stats. I _think_ these will be more accurate and less confusing for newcomers. I ran and re-ran with the seed, my best guess is the previous figures came from a slightly different config / data. 👍

sambrilleman:
Thanks for spotting this @statwonk. And even bigger thanks for going to the hassle of resolving it! 👍

PR #327, opened 2019-01-16
stan_jm: possible copy-paste error in dokumentation

snaut:
Two options were described as default in the basehaz parameter, I corrected this to the one fitting the function definition and description in the details section.


sambrilleman:
Thanks! Looks good to me.

PR #338, opened 2019-03-07
fix kfold to work when folds given and K!=10

avehtari:
Fix kfold to work if folds are given but the number of folds in folds is different from 10 (the default value for K in kfold function)


jgabry:
Thanks Aki!

PR #340, opened 2019-03-16
loo and kfold updates

jgabry:
* `loo_compare` replaces the now deprecated `compare_models` as disccused in https://github.com/stan-dev/loo/pull/93

* `kfold` now parallelizes by fold instead of by Markov chain (unless otherwise specified) and has its own documentation page

See my comment below for more details on the changes and example code.


jgabry:
## More details on the changes

#### loo_compare 

There are three ways rstanarm users can use `loo_compare`:

```r
# 1) By passing loo, waic, or kfold objects directly to the 
# default loo_compare() method in the loo package
loo1 <- loo(fit1)
loo2 <- loo(fit2)
loo_compare(loo1, loo2) # calls loo::loo_compare.default

# 2) By storing loo, waic, or kfold objects in the corresponding stanreg objects 
# and then passing the stanreg objects to the loo_compare.stanreg method in rstanarm.
# This allows rstanarm to validate that the models are suitable for comparison
fit1$loo <- loo1
fit2$loo <- loo2
loo_compare(fit1, fit2) # calls rstanarm::loo_compare.stanreg

# 3) By putting the stanreg objects with 'loo' components into a stanreg_list
# and then passing the stanreg_list to the loo_compare.stanreg_list method in rstanarm. 
# This also allows rstanarm to validate that the models are suitable for comparison.
models <- stanreg_list(fit1, fit2)
loo_compare(models) # calls rstanarm::loo_compare.stanreg_list
```

### k-fold parallelization

`kfold` now parallelizes by fold instead of by Markov chain (unless otherwise specified) and has its own documentation page. The new `cores` argument to `kfold` can be used as follows:

```r
# Example code demonstrating the different ways to specify the number 
# of cores and how the cores are used

# starting with no mc.cores set
options(mc.cores = NULL)
 
# spread the K models over N_CORES cores (method 1, cores argument)
kfold(fit, K, cores = N_CORES)

# spread the K models over N_CORES cores (method 2, mc.cores option)
options(mc.cores = N_CORES)
kfold(fit, K)

# fit K models sequentially using N_CORES cores for the Markov chains each time
options(mc.cores = N_CORES)
kfold(fit, K, cores = 1)
```
On Windows it uses `parallel::parLapply` with a PSOCK cluster, and for Mac and Linux it uses `parallel::mclapply`.

avehtari:
I would like the following to work
```
kfold_1 <- kfold(fit_1, K=10)
loo_2 <- loo(fit_2)
loo_compare(kfold_1, loo_2)
```
This is useful if `loo(fit_1)` has many high k values. K-fold-cv and loo are not exactly same unless K=N, but it's still a better comparison than comparing loo with many high k's.

avehtari:
Other examples from https://avehtari.github.io/RAOS-Examples/Mesquite/mesquite.html

1.

``fit_1 <- stan_glm(weight ~ diam1 + diam2 + canopy_height + total_height +
                      density + group, data=mesquite, seed=SEED, refresh=0)
print(fit_1)
(loo_1 <- loo(fit_1))
fit_2 <- stan_glm(log(weight) ~ log(diam1) + log(diam2) + log(canopy_height) +
                      log(total_height) + log(density) + group,
                  data=mesquite, seed=SEED, refresh=0)
(loo_2 <- loo(fit_2))
```
So these models are not directly comparable as the target is different.
`loo_compare(loo_1, loo_2)` happily compares them without a warning.
however these objects have necessary information as 
`compare(loo_1, loo_2)` can recognize different yhash'es
```
Attributes: < Component “formula”: formulas differ in contents >
Attributes: < Component “name”: 1 string mismatch >
Attributes: < Component “yhash”: 1 string mismatch >
```
It would be nice if `loo_compare` would also check if there is yhash attribute and give a warning.

2.
In that example, I also make the necessary Jacobian correction
```
#' **Jacobian correction to make the models comparable**<br>
#' Jacobian correction is needed as model 1 is modeling y and model 2
#' is modeling log(y).
loo_2_with_jacobian <- loo_2
loo_2_with_jacobian$pointwise[,1] <- loo_2_with_jacobian$pointwise[,1]-
                                     log(mesquite$weight)
(elpd_loo_2_with_jacobian <- sum(loo_2_with_jacobian$pointwise[,1]))
#' `compare_models` checks that the target data is same, but `compare`
#' trusts that we know what we are comparing
loo_compare(loo_1, loo_2_with_jacobian)
```
This is the reason that I would give just a warning if loo objects have mismatch in yhash, as the user might be doing the Jacobian correction like this. Suggestions for making this better?

**Comment by avehtari** [comment]
otherwise looks great!

**Comment by jgabry** [comment]
Thanks Aki. To get the`loo_compare.default` to check for the y hashes will require a small change in the loo package (we can do that next release, and I agree a warning is fine instead of an error), but everything else you suggested should now be possible.

**Comment by avehtari** [comment]
Great, I'll make some testing tomorrow.

**Comment by avehtari** [comment]
After additional thinking and testing, it would useful to allow but give a
warning for `loo_compare(kfold_1, loo_2)` and `loo_compare(kfold_1, kfold_2)` if K is not the same and for loo K=n. Warning like "Warning: comparing 10-fold-CV to loo. For more accurate comparison use same number of folds or loo for all models compared."  (in the second case 10-fold-CV to 20-fold-CV)

**Comment by avehtari** [comment]
Parallel kfold is great!

**Comment by jgabry** [comment]
Thanks Aki. I can have the loo package throw those warnings instead of rstanarm, that way it will apply to brms too, etc. I'll get that in the next loo release. (https://github.com/stan-dev/loo/pull/103)

@bgoodri I think this is probably ready to go.

**Comment by avehtari** [comment]
Thanks, Jonah. Ok, for me.

**Comment by avehtari** [comment]
On mobile until tomorrow, so I can't make an official review approval, but go ahead with this

**Comment by bgoodri** [comment]
Why are kfold.stanreg loo_compare.stanreg waic.stanreg being exported? There are warnings about them
```
* checking S3 generic/method consistency ... NOTE
Found the following apparent S3 methods exported but not registered:
  kfold.stanreg loo_compare.stanreg waic.stanreg
See section ‘Registering S3 methods’ in the ‘Writing R Extensions’
manual.
```

jgabry:
I may have made a mistake with the roxygen commands. I can check/fix tomorrow.

jgabry:
Should be fixed by 1527cf17

PR #343, opened 2019-03-20
bugfix/stanjm assoc data interaction

### LINKED ISSUE #342: stan_jm: bug in *_data association structures
#### Summary:

In `stan_jm`, association structures of the form `etavalue_data` and `etaslope_data` are not being estimated correctly. There is a small bug in the indexing in one of the .stan files, that causes the estimates for those association parameters to go seriously wrong. 

I've found the bug, but I'm still opening this issue to track the pull request that fixes it.

This was highlighted in a [post](https://discourse.mc-stan.org/t/joint-model-yields-inflated-association/8003/4) on the Stan Forums.

#### Reproducible Steps:

```
f1 <- stan_jm(formulaLong = logBili ~ year + (1 | id), 
              dataLong = pbcLong,
              formulaEvent = Surv(futimeYears, death) ~ trt, 
              dataEvent = pbcSurv,
              time_var = "year",
              assoc = c("etavalue", "etavalue_data(~ trt)"),
              chains = 1, seed = 12345, iter = 1000)
```

#### RStanARM Version:

<= 2.18.2

#### R Version:

NA

#### Operating System:

NA

sambrilleman:
Closes #342. This fixes a serious bug in the `etavalue_data` and `etaslope_data` association structures for `stan_jm`.

### DIFF
--- a/src/stan_files/model/assoc_evaluate.stan
+++ b/src/stan_files/model/assoc_evaluate.stan
@@ -37,7 +37,6 @@
           mark += 1;
           e_eta_q += a_beta[mark] * (val - a_xbar[mark]);
         }
-        mark2 += 1; // count even if assoc type isn't used
         if (has_assoc[9,m] == 1) { // etavalue*data
           int J = a_K_data[mark2];
           int j_shift = (mark2 == 1) ? 0 : sum(a_K_data[1:(mark2-1)]);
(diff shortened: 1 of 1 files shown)


bgoodri:
OK, we'll get this into the imminent release of rstanarm and try to do the
other survival stuff for the following release. A rather large refactoring
of the build process is needed if we are going to be able to run R CMD
check on Windows in less than 10 minutes.

sambrilleman:
Ok, thanks. 

Re the survival models, what if I run `stan_mvmer` through jm.stan instead of through mvmer.stan (from memory they use the same or similar chunks, so with a bit of refactoring I think I can combine them) -- that would potentially allow us to delete mvmer.stan and then replace it with a surv.stan file instead. The total number of .stan files would remain the same in **rstanarm**, so compiling time shouldn't change much? Is this an easier solution, compared with refactoring the entire build process?

bgoodri:
We may have to do some of both but let's figure it out after 2.19.

PR #344, opened 2019-03-22
Optimizing diagnostics and importance resampling after optimizing

### LINKED ISSUE #334: log_p and log_g missing from object returned by stan_glm(..., algorithm='optimizing')
#### Summary:
The object returned by rstan `optimizing()`, includes `log_p` and `log_g`, which are needed for Pareto k diagonostic of Laplace approximation and loo. The object returned by rstanarm `stan_glm(..., algorithm='optimizing')` is missing these. I did check `stan_glm.fit.R`, but could not figure where these are lost.

Fix: pass `log_p` and `log_g` from `optimizing()` to the object returned by `stan_glm()`
I guess the same holds for `stan_betareg()`

#### Reproducible Steps:

The following works
```
library(brms)
library(rstan)
x <- 1:20
n <- length(x)
a <- 0.2
b <- 0.3
sigma <- 0.5
y <- a + b*x + sigma*rnorm(n)
fake <- data.frame(x, y)
sc <- make_stancode(y ~ x, data = fake)
sd <- make_standata(y ~ x, data = fake)
sm <- stan_model(model_code=sc)
so <- optimizing(sm, data=sd, hessian = TRUE, draws = 4000)
length(so$log_p)
```
The following is missing `log_p` and `log_g`
```
library(rstanarm)
fit <- stan_glm(y ~ x, data = fake, algorithm = 'optimizing', draws = 4000)
length(fit$log_p)
```

#### RStanARM Version:
2.18.2

Edit: added missing line to the code example

avehtari:
- Fixes #334
- adds Pareto-k diagnostic for optimizing
- enables importance resampling by default, so that remaining draws have equal weights
- enables Monte Carlo diagnostics in summary()
- enables posterior interval, posterior_predict
- enables loo and loo_predict


R/stan_betareg.fit.R
```diff
@@ -35,7 +35,7 @@ stan_betareg.fit <-
            algorithm = c("sampling", "optimizing", "meanfield", "fullrank"),
            adapt_delta = NULL, 
            QR = FALSE,
-           draws = 1000) {
+           draws = 4000) {
```
bgoodri: Is there a reason why this is 4000 for stan_betareg but not stan_glm?

avehtari: Sorry, I was going between more accurate 4000 and faster 1000 (with faster log density computations 4000 would be better for importance sampling, but with slower computations we get some result much faster and then diagnostics recommend to increase draws for more accuracy). I switched this back to 1000. Furthermore, diagnostics and resampling is not yet enabled for betareg.

avehtari:
To clarify status of stan_betareg. With this PR, log_p and log_g are stored when using stan_betareg or stan_glm, but optimizing diagnostics is enabled only for stan_glm.

avehtari:
rebased onto master

avehtari:
rebased onto master

avehtari:
Not certain if this works anymore after the rebase. The way Stan is called has changed to
`out <- do.call(optimizing, args = optimizing_args)` which causes now error. Looking into it, but too late today.

bgoodri:
I thought it would be too confusing to have arguments like `draws` and especially `thin` that could be misinterpreted as pertaining to NUTS, so I tried to get them out of the ... when specified and then used your defaults when they were unspecified. It should be fixed now if you pull.

avehtari:
Did not help. And the error happens inside `optimizing` which is in rstan.

avehtari:
Ah, I had forgot that rstan fails doing draws from the posterior of one 1 parameter model. I had noticed that before, but didn't report or fix as sampling is fast for 1 parameter models, but no wanted something quick to test. This PR works now (for models with more than 1 parameter).

PR #350, opened 2019-04-17
pass and check importance_resampling option

avehtari:
- passes importance_resampling option to rstan
- computes Pareto diagnostic and resampling only if enabled

this will make it possible to avoid possibly costly log_p computation for approximate posterior draws


avehtari:
This has been approved but I'm not authorized to merge this. Without this stan_glm optimizing fails.

jgabry:
Ok merged now!

PR #357, opened 2019-05-08
do OLS if K_smooth > 0

avehtari:
OLS trick works also for stan_gamm4 with smooths when `len_y >= (has_intercept + K + K_smooth)`. `S` is appended to `X_` and `beta_smooth` is appended to coeff vector, and that's all! Significant speedup is obtained when applicable.

### DIFF
--- a/src/stan_files/continuous.stan
+++ b/src/stan_files/continuous.stan
@@ -140,7 +140,9 @@ transformed parameters {
 }
 model {
   if (can_do_OLS) {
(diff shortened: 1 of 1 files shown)


jgabry:
Cool

avehtari:
For example, 4 times faster (~40s -> ~10s) with n=827 and
```
fits <- stan_gamm4(doy ~ s(year), data=cherry_doy, seed=583829, cores=4)
```
The speed difference is likely to be bigger with bigger n.

PR #360, opened 2019-05-16
speed up with bernoulli_logit_glm when applicable

avehtari:
Use bernoulli_logit_glm_lpmf when applicable.
E.g. the following examples are 4 times faster (same with n=1e3, n=1e5, and n=1e6).

```
#' **Logistic regression**<br>
#' **Create fake data with 10\,000 observations and p=100**
SEED <- 1655
set.seed(SEED)
n <- 1e4
x <- rnorm(n)
xn <- matrix(rnorm(n*999),nrow=n)
a <- 2
b <- 3
sigma <- 1
y <- (a + b*x + sigma*rnorm(n))>0+0
fake <- data.frame(x, xn, y)

#' **Fit and display using stan_glm and MCMC**<br>
fit2 <- stan_glm(y ~ ., data=fake, family=binomial(), mean_PPD=FALSE,
                 init_r=0.1, seed=SEED)

#' **Fit and display using stan_glm and optimization**<br>
fit3 <- stan_glm(y ~ ., data=fake, family=binomial(),
                algorithm='optimizing', init=0)
```

### DIFF
--- a/src/stan_files/bernoulli.stan
+++ b/src/stan_files/bernoulli.stan
@@ -75,6 +75,11 @@ transformed data {
   int<lower=0> successes[clogit ? J : 0];
   int<lower=0> failures[clogit ? J : 0];
   int<lower=0> observations[clogit ? J : 0];
+
+  int can_do_bernoullilogitglm = link == 1 && clogit == 0 && has_offset == 0 && prior_PD == 0 && dense_X == 1 && has_weights == 0 && t == 0;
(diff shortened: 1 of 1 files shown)


bgoodri:
We can do this one. For the version of rstanarm after the next one, I think we need to write our own C++ functions that handle the derivatives so that they can work in cases where there is a `Z` matrix, etc.

PR #379, opened 2019-08-01
Set the random seed before calling rexp() in loo_R2()

### LINKED ISSUE #378: loo_r2 results not reproducible between calls
#### Summary:
Two calls to `loo_r2` on the same fitted object don't produce the same result.

#### Description:
Inside `loo_r2` there's a call to `rexp()` to generate the Dirichlet weights. Since no seed is set before calling it, results are not exactly reproducible between calls.

#### Reproducible Steps:
```
fit <- stan_glm(mpg ~ wt, data = mtcars, seed = 100)
all.equal(loo_R2(fit), loo_R2(fit))
# [1] "Mean relative difference: 0.1023236"
```

#### RStanARM Version:
Current github

#### R Version:
3.6.1

#### Operating System:
Ubuntu Linux 18.04.2 LTS

mcol:
This allows two calls to loo_R2() on the same fitted object to produce identical results, without affecting the state of the RNG outside of the function. This fixes #378.

### DIFF
--- a/R/bayes_R2.R
+++ b/R/bayes_R2.R
@@ -110,7 +110,13 @@ loo_R2.stanreg <- function(object, ...) {
   
   S <- nrow(mu_pred)
   N <- ncol(mu_pred)
-  
+
+  # set the random seed as the seed used in the first chain and ensure
+  # the old RNG state is restored on exit
+  rng_state_old <- .Random.seed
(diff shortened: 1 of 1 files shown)


jgabry:
This is great, thanks!

PR #409, opened 2020-01-04
Adding ab-testing vignette

imadmali:
I'd like to contribute a vignette that outlines how to do A/B testing inference (post-experiment) with rstanarm functions. As discussed [here](https://github.com/stan-dev/stan-dev.github.io/pull/97) I think it would be useful if it was part of the "Additional tutorials on specific modeling techniques and applications" section on the [vignette page](https://mc-stan.org/rstanarm/articles/index.html) of the rstanarm site. @jgabry I've left your comments in and have changed the content based on some of them. Also, regarding _Appendix B_, I don't fully understand how priors are specified on group specific intercepts when using `stan_glmer` so I just said something to the effect of "default priors are used".

### DIFF
--- a/vignettes/ab-testing.Rmd
+++ b/vignettes/ab-testing.Rmd
@@ -459,8 +459,9 @@ compute the overlap proportion of these two sets of predictions.
 # below doesn't work as expected (predictions are bigger than the number of trials)
 # pp_c <- posterior_predict(fit_group_bin, newdata = data.frame(group = factor("C"), trials = 19))
 # pp_d <- posterior_predict(fit_group_bin, newdata = data.frame(group = factor("D"), trials = 22))
-pp_c <- posterior_predict(fit_group_bin)[,1]
-pp_d <- posterior_predict(fit_group_bin)[,2]
+pp <- posterior_predict(fit_group_bin)
+pp_c <- pp[,1]
+pp_d <- pp[,2]
 pp_c_quant <- quantile(pp_c, probs = c(0.05,0.95))
 pp_d_quant <- quantile(pp_d, probs = c(0.05,0.95))
 
--- a/vignettes/ab-testing.Rmd
+++ b/vignettes/ab-testing.Rmd
@@ -1,0 +1,479 @@
+---
+title: "Probabilistic A/B Testing with rstanarm"
+author: "Imad Ali"
+date: "`r Sys.Date()`"
+output:
+  html_vignette:
+    toc: yes
+params:
+  EVAL: !r identical(Sys.getenv("NOT_CRAN"), "true")
+bibliography: mrp-files/mrp.bib
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE)
+knitr::opts_chunk$set(collapse = TRUE)
+library(rstan)
+library(rstanarm)
+library(dplyr)
+rstan_options(auto_write = TRUE)
+```
+
+## Abstract
+
+This case study shows how basic A/B testing using Stan and Bayesian methods can
+be used to facilitate business decisions. In practice, we find this approach
+useful given its ability to quantify domain-specific business knowledge and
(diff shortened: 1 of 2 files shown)


vignettes/ab-testing.Rmd
```diff
@@ -0,0 +1,479 @@
...
+    toc: yes
+params:
+  EVAL: !r identical(Sys.getenv("NOT_CRAN"), "true")
+bibliography: mrp-files/mrp.bib
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE)
+knitr::opts_chunk$set(collapse = TRUE)
+library(rstan)
```
jgabry: Is rstan actually used anywhere anymore?

vignettes/ab-testing.Rmd
```diff
@@ -0,0 +1,479 @@
...
+library(rstanarm)
+library(dplyr)
```
jgabry: Is it possible to do this without dplyr? I don't think we currently have dplyr as a dependency but we would need to add it to use it in the vignette.

imadmali:
@jonah regarding the hierarchical section I'm fine with saying default priors if there is a place where the information about the priors is clearly defined? If there is such a place then I'll provide a link to it. Let me know.

vignettes/ab-testing.Rmd
imadmali: So I use dplyr [here](https://github.com/stan-dev/rstanarm/blob/doc/ab-testing/vignettes/ab-testing.Rmd#L408-L409). I don't have to use it and it's straightforward to rewrite this in non-dplyr code. But since it's pretty much a standard for industry workflow maybe it's something to consider using in all the vignettes. But then, like you mentioned, it's not a dependency for the main purpose of the package. Let me know what you want me to do.

imadmali:
@jgabry I made some more edits. Let me know if it's okay to merge or if more changes need to be made.

vignettes/ab-testing.Rmd
```diff
@@ -0,0 +1,739 @@
+---
+title: "Probabilistic A/B Testing with rstanarm"
+author: "Imad Ali"
+date: "`r Sys.Date()`"
+output:
+  html_vignette:
+    toc: yes
+params:
+  EVAL: !r identical(Sys.getenv("NOT_CRAN"), "true")
```
bgoodri, line 9: We can't use the `params` mechanism any more because some OS does not have a sufficiently recent Pandoc or something. The way we are doing the other vignettes now is to put `knitr::opts_chunk$set(eval = identical(Sys.getenv("NOT_CRAN"), "true")))`

ssp3nc3r:
Nice vignette. Would it make sense to call `set.seed()` for repeated calls to `posterior_predict` when comparing posteriors, or just run the function once and subtract the groups? For example,

```
pp_c <- posterior_predict(fit_group_bin)[,1]
pp_d <- posterior_predict(fit_group_bin)[,2]
```

would be

```
pp <- posterior_predict(fit_group_bin)
pp_c <- pp[,1]
pp_d <- pp[,2]
```

vignettes/ab-testing.Rmd
jgabry: I got around this in other packages just by checking `if (exists("params"))` but not using it is also fine.

jgabry:
@ssp3nc3r Yeah good suggestion.

vignettes/ab-testing.Rmd
imadmali: Ok, I'll just remove it then.

imadmali:
I've made all suggested edits. Let me know if I'm missing anything.

imadmali:
@jgabry @bgoodri is there anything left that I need to do to get this accepted?

jgabry (approved):
I just went through and removed `autoscale=FALSE` since it's not needed anymore with the latest release. AFAIK that was the last remaining thing to change so will merge this.

PR #421, opened 2020-02-23
Update posterior_survfit.R

liuxiangyu127:
also output the sampled survival probabilities

### DIFF
--- a/R/posterior_survfit.R
+++ b/R/posterior_survfit.R
@@ -513,7 +513,7 @@ posterior_survfit <- function(object, newdataLong = NULL, newdataEvent = NULL,
             last_time = last_time, ids = id_list, draws = draws, seed = seed, 
             offset = offset)
   if (dynamic && !is.null(newdataEvent)) {
-    out <- structure(out, b_new = b_new, acceptance_rate = acceptance_rate)
+    out <- structure(out, surv = surv, b_new = b_new, acceptance_rate = acceptance_rate)
   }
   out
 }


sambrilleman:
@jgabry @liuxiangyu127 Seems reasonable, but should we maybe add an argument to make this optional? It might dramatically inflate the size of the return object, so probably only want to return it when people explicitly ask for it.

In `posterior_traj` I have a similar thing, but I called the argument `return_matrix`, see [here](https://github.com/stan-dev/rstanarm/blob/140246d65dd02e8ad4558c640080e88fe537068f/R/posterior_traj.R#L146). Perhaps we could call it the same thing here? The minor difference is that in `posterior_traj` with `return_matrix = TRUE` it is **only** the matrix of draws that is returned, whereas here it seems @liuxiangyu127 wants to return the matrix of draws in addition to all the other stuff, which is probably fine, but we'd just want to be aware that doing that will mean `posterior_traj` and `posterior_survfit` would behave slightly differently for `return_matrix = TRUE`.

@liuxiangyu127 @jgabry -- would be happy for either of you to make these changes if you want. I don't have much spare time for this stuff atm! (hence my slow reply, sorry!! 😦 )

bgoodri:
Is this tantamount to a `posterior_epred` method for survival models?

sambrilleman:
Yeah it may well be. I don't think `posterior_epred()` existed back when `posterior_survfit()` came into existence. So I will read the more recent docs and see.

sambrilleman:
Had a look and there are some differences.

`posterior_epred` is kinda like `posterior_survfit(type = "haz")` but as soon as you start looking for `posterior_survfit(type = "surv")` you need a transformation of the linear predictor that isn't the link function but rather the relationship between the survival probability and the hazard (since the linear predictor is an estimate of the log hazard).

Also there isn't always a closed form for the survival probability, so depending on the model (e.g. if the model has time-varying coefficients) then `posterior_survfit(type = "surv")` might use quadrature.

By default `posterior_survfit()` also does a bit of summarising into median and interval bounds for the predictions, rather than returning the entire DxN matrix of posterior draws (this is because the time component in the prediction data would make the DxN matrix super difficult for the user to interpret unless they really want it).

PR #434, opened 2020-05-03
Fix #431: Make drop_redundant_dims safe for data.table

### LINKED ISSUE #431: drop_rendudant_dims() issues error with data.table
#### Summary:

When a data.table is provided as data, the internal function drop_redundant_dims() results in error.

#### Description:

When a data.table is provided to rstanarm as data, the following error is produced:

```
Called from: stop("j (the 2nd argument inside [...]) is a single symbol but column name '", 
    jsubChar, "' is not found. Perhaps you intended DT[, ..", 
    jsubChar, "]. This difference to data.frame is deliberate and explained in FAQ 1.1.")
```
This is because the method dispatch for `[` is finding the method for data.table before data.frame.

Since the error message states this is intentional for data.tables, a check and/or coercion to data.frame might be warranted.

#### Reproducible Steps:
```
library(rstanarm)
library(data.table)
mt = as.data.table(mtcars)
stan_glm(mpg ~ cyl, data = mt)
```

#### RStanARM Version:

2.19.3

#### R Version:

3.6.3

#### Operating System:

Arch Linux 5.6.5-1

danschrage:
Modify `drop_redundant_dims` to avoid an error when data is a data.table, as reported in #431. This happens because `drop_redundant_dims` indexes into a data.frame using a vector of logicals, but data.table expects an unquoted name to be a column name. To fix this, my code skips the dimension-reduction step if data is a data.table.

This shouldn't cause problems with data.table: In general, dimension reduction like this isn't necessary for a data.table, because data.table makes it really difficult to add a matrix as a column. The only way to do this is to explicitly coerce an existing data.frame that has matrices as columns using `setDT()`, and that will warn the user against doing this (see https://github.com/Rdatatable/data.table/pull/3851). For any other case, data.table will automatically coerce matrices into columns. In other words, data.table does this dimension reduction automatically, so it can be safely skipped.

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -743,6 +743,9 @@ check_stanfit <- function(x) {
 #   otherwise if_missing is returned.
 #
 drop_redundant_dims <- function(data) {
+  if(inherits(data, "data.table")) {
+    return(data)
+  }
   drop_dim <- sapply(data, function(v) is.matrix(v) && NCOL(v) == 1)
   data[, drop_dim] <- lapply(data[, drop_dim, drop=FALSE], drop)
   return(data)


mespe:
I am not sure this is the best solution for this issue. Specifically, I worry it is not robust because it relies on implicit behavior of the data.table package. It would be more robust to either issue a more informative error and let the user coerce the input data, or force coercion to a data.frame. 

Additionally, it would be nice if this PR included an associated test so if/when data.table changes its behavior, it will flag this as needing attention.

danschrage:
Thanks for the comments! I updated the pull request, and now `validate_data` behaves identically to `validate_newdata` and simply coerces anything (data.table, tibble, etc.) into a data.frame before calling `drop_redundant_dims`. So it should no longer depend on or be disrupted by any future changes in data.table. That should eliminate the need for an associated test, too.

I tested it on your MWE from #431 as well as on my own data.

jgabry:
@danschrage and @mespe, thanks for sorting this out!

PR #456, opened 2020-07-31
fix log-lik for stan_polr

### LINKED ISSUE #455: log_lik() with newdata is sometimes wrong for stan_polr()
@bgoodri There's a few spots where I think `log_lik()` is doing the wrong thing for `stan_polr()` models, but I'm not 100% sure of the fix, so I have a few questions for you.

For `stan_polr()`, if `newdata` is specified to `log_lik()` (which it is when doing reloo or kfold), then errors can occur here

https://github.com/stan-dev/rstanarm/blob/709d0db9000fb53b15f0c8f93aef755e18c04209/R/log_lik.R#L323-L326

because `y` is referring to `y` from `newdata`, which might only have two unique values while the original y has more than 2 levels. In that case `stanmat[, zetas, drop = FALSE]` doesn't find the right columns because `patt` is wrong (`patt` thinks it's a binary model when it isn't). 

So, a few questions: 

1. Should we just always define `patt`

https://github.com/stan-dev/rstanarm/blob/709d0db9000fb53b15f0c8f93aef755e18c04209/R/log_lik.R#L323

using the original `y` and not the `y` from `newdata`? 

2. Should `max_y` also always be defined using the original data and not newdata? This is important because in `ll_polr_i` does `J <- draws$max_y` and then uses `J` when computing the log-lik:

https://github.com/stan-dev/rstanarm/blob/709d0db9000fb53b15f0c8f93aef755e18c04209/R/log_lik.R#L463-L469

If you want to actually see this bug in action, first pull the recent commits to master (this bug isn't apparent until you pull in changes fixing another error that would happen first) and then run the `stan_polr()` model provided in this discourse post (data provided in post): https://discourse.mc-stan.org/t/loo-with-k-threshold-error-for-stan-polr/17052/6

jgabry:
fixes #455

@bgoodri I think this is correct but I'll make a few review comments to indicate the reason for certain changes. Let me know if you seen any issues.

### DIFF
--- a/R/log_lik.R
+++ b/R/log_lik.R
@@ -320,8 +320,7 @@ ll_args.stanreg <- function(object, newdata = NULL, offset = NULL, m = NULL,
     }
     data <- data.frame(y, x)
     draws$beta <- stanmat[, colnames(x), drop = FALSE]
-    patt <- if (length(unique(y)) == 2L) "(Intercept)" else "|"
-    zetas <- grep(patt, colnames(stanmat), fixed = TRUE, value = TRUE)
+    zetas <- grep("|", colnames(stanmat), fixed = TRUE, value = TRUE)
     draws$zeta <- stanmat[, zetas, drop = FALSE]
     draws$max_y <- max(y)
     if ("alpha" %in% colnames(stanmat)) { 
@@ -456,8 +455,8 @@ ll_args.stanreg <- function(object, newdata = NULL, offset = NULL, m = NULL,
 .ll_polr_i <- function(data_i, draws) {
   eta <- linear_predictor(draws$beta, .xdata(data_i), data_i$offset)
   f <- draws$f
-  J <- draws$max_y
   y_i <- data_i$y
+  J <- ncol(draws$zeta) + 1
   linkinv <- polr_linkinv(f)
   if (is.null(draws$alpha)) {
     if (y_i == 1) {
--- a/R/stan_polr.R
+++ b/R/stan_polr.R
@@ -141,6 +141,11 @@ stan_polr <- function(formula, data, weights, ..., subset,
                       do_residuals = NULL) {
 
   data <- validate_data(data, if_missing = environment(formula))
+  is_char <- which(sapply(data, is.character))
+  for (j in is_char) {
+    data[[j]] <- as.factor(data[[j]])
+  }
+  
   algorithm <- match.arg(algorithm)
   if (is.null(do_residuals)) 
     do_residuals <- algorithm == "sampling"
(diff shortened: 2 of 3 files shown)


R/log_lik.R
```diff
@@ -320,8 +320,7 @@ ll_args.stanreg <- function(object, newdata = NULL, offset = NULL, m = NULL,
     }
     data <- data.frame(y, x)
     draws$beta <- stanmat[, colnames(x), drop = FALSE]
-    patt <- if (length(unique(y)) == 2L) "(Intercept)" else "|"
-    zetas <- grep(patt, colnames(stanmat), fixed = TRUE, value = TRUE)
+    zetas <- grep("|", colnames(stanmat), fixed = TRUE, value = TRUE)
```
jgabry, line 323: I think we can get rid of `patt` here because length(unique(y)) can never legitimately be 2 if `is_polr(object)` is TRUE. When you fit a binary model with `stan_polr()`, `is_polr(object)` is FALSE, so this part of the `log_lik()` code isn't used.

R/log_lik.R
```diff
@@ -456,8 +455,8 @@ ll_args.stanreg <- function(object, newdata = NULL, offset = NULL, m = NULL,
 .ll_polr_i <- function(data_i, draws) {
   eta <- linear_predictor(draws$beta, .xdata(data_i), data_i$offset)
   f <- draws$f
-  J <- draws$max_y
   y_i <- data_i$y
+  J <- ncol(draws$zeta) + 1
```
jgabry, line 459: Don't use `max_y` since `max_y` can change if using `newdata`.

R/stan_polr.R
```diff
@@ -141,6 +141,11 @@ stan_polr <- function(formula, data, weights, ..., subset,
                       do_residuals = NULL) {
 
   data <- validate_data(data, if_missing = environment(formula))
+  is_char <- which(sapply(data, is.character))
+  for (j in is_char) {
+    data[[j]] <- as.factor(data[[j]])
+  }
+
```
jgabry, line 148: Need to convert any character predictors to factors to avoid a potential error in model.matrix() 

```
Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) : 
  contrasts can be applied only to factors with 2 or more levels
```

which happens when `newdata` is specified for reloo or kfold and a character predictor that had multiple unique values only has 1 in `newdata`. Converting to factor immediately when fitting initial model preserves the other levels even if they're not present in `newdata`.

R/log_lik.R
bgoodri: OK

R/log_lik.R
bgoodri: Also, Andrew made us put in support for estimating more categories than there are in the original outcome anyway.

R/stan_polr.R
bgoodri: OK, whatever

PR #465, opened 2020-08-26
Vignette prior predictive

lauken13:
Hi Jonah,

Just wanted to follow up with this vignette, is it still useful?


jgabry:
Thanks for reminding me! Somehow this wasn't on my to-do list anymore but should have been. I'll try to spend a bit of time on it ASAP. I think we should include it in the next rstanarm release.

jgabry:
Is it ok if I make edits/additions directly or do you prefer that I just make comments?

lauken13:
Nah go ahead and make edits, also add yourself as a co-author! :)

Lauren

jgabry:
Ok cool will do!

jgabry:
@lauken13 I made a bunch of edits to the first half (it looks like I changed more than I actually did because some changes are just reflowing lines). I'll go through the second half (the part about priors for the slope) soon. 

A few general things that I noticed: 

* Because rstanarm doesn't list dplyr or tidyr in the DESCRIPTION file we'll need to either not use dplyr and tidyr or do something similar to the MRP vignette where we pre-processed anything that needed those packages. Alternatively we can add them to the `Suggests` field in the DESCRIPTION file if @bgoodri is OK with it. 

* We might want to consider adding another example that isn't logistic regression. It occurred to me that logistic regression is actually a special case where it makes more sense to look at the latent Pr(y = 1) instead of actual values of y simulated from the bernoulli using Pr(y = 1). Because we don't actually simulate outcomes this example technically doesn't use the prior predictive distribution, which could be confusing. I tried to mention this in some of my edits, but it's still a bit weird not to have an actual example of drawing from the predictive distribution. What do you think?

lauken13:
Yeah I think that sounds like a good idea. Do you have any suggestions for
other examples? Normal linear regression?

I'd be fine to redo the data manipulation with base R, if that makes more
sense for the package!

jgabry:
> Yeah I think that sounds like a good idea. Do you have any suggestions for other examples? Normal linear regression?

Yeah I think linear regression would good. I don't think we need anything too complicated, just something to illustrate using `posterior_predict()` instead of `posterior_epred()`. 

> I'd be fine to redo the data manipulation with base R, if that makes more sense for the package!

If that's not too much of a pain that would be great!

PR #474, opened 2020-10-08
track manual pages

### LINKED ISSUE #466: Start tracking Rd files on GitHub
rstanarm doesn't track Rd files on GitHub, which results in the doc not being included in various situations (see e.g. @ermeel86's comment https://github.com/stan-dev/rstanarm/issues/442#issuecomment-683775455, but there are also other cases) . @bgoodri I forget why this decision was made initially, but rstanarm might be the only R package on GitHub that doesn't track the Rd files (at least it's hard to find another one). I don't think there's a good reason not to track them at this point, so I think we should start doing it. Objections?

jgabry:
Fixes #466

This PR starts tracking the manual pages. rstanarm might be the only R package on GitHub that doesn't track the Rd files (at least it's hard to find another one). I don't think there's a good reason not to track them at this point and it's causing problems when installing rstanarm from github. 

@bgoodri if you're ok with this then since it affects whether people get the doc when installing from github it would be good to merge this asap rather than waiting for the next release. 

@sambrilleman @ermeel86 Once this is merged can you update the feature/survival branch too so we can fix the problem of missing doc when installing from there?


bgoodri:
This is a recipe for having documentation that is out of date with what the code does. If you want people to be able to install from GitHub and be able to read the manual pages, the uncomment those lines from `cleanup` and `cleanup.win`. We commented those out a long time ago when roxygen2 went into an infinite loop but roxygen2 is better now.

ermeel86:
> @sambrilleman @ermeel86 Once this is merged can you update the feature/survival branch too so we can fix the problem of missing doc when installing from there?

Happy to do so.

jgabry:
> This is a recipe for having documentation that is out of date with what the code does.

I get your point but I don't think it's a big concern. This hasn't been a problem with any of our other R packages, so I'm not sure why it would be a problem for rstanarm.  I'm in the habit of roxygenizing before committing, but it's also pretty easy to set up commit hooks to make sure things don't get out of sync. Regardless, even slightly out of sync doc is much better than no doc. 

> If you want people to be able to install from GitHub and be able to read the manual pages, the uncomment those lines from cleanup and cleanup.win. We commented those out a long time ago when roxygen2 went into an infinite loop but roxygen2 is better now.

I think it was a mistake to comment those out without tracking the manual pages. Getting the doc when installing a package is essential. If roxygen2 won't go into an infinite loop now then I'm fine going with that approach again, but if it stops working again then I think we absolutely need to track the manual pages. I don't think the doc will get too out of sync, but even if it does a bit, that's much better than no doc at all.

jgabry:
I really still think we should be tracking the manual pages like every other R package does, but I will close this since we have #477 as an alternative. If #477 isn't a reliable alternative then I think we have to start tracking the manual pages because we can't have people install this without documentation.

PR #477, opened 2020-10-15
uncomment roxygen lines from cleanup[.win]

jgabry:
This is an alternative to #474 that adds roxygenize() to cleanup and cleanup.win instead of tracking manual pages.

### DIFF
--- a/cleanup
+++ b/cleanup
@@ -1,5 +1,5 @@
 #!/bin/sh -e
 
 # Note to Windows users: This is not actually platform specific.
-# "${R_HOME}/bin/R" --vanilla --slave -e 'roxygen2::roxygenize(clean = FALSE)'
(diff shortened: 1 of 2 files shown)


sambrilleman:
Yeah cool, if this works then great! I guess any solution that works for both `install_github()` and the built binaries...

jgabry:
Yeah hopefully this works for both, but if not then we should start tracking the manual pages. 

Once we merge this we can rebuild the windows binary and also make a mac one.

jgabry:
@bgoodri Is this no longer necessary after your commit https://github.com/stan-dev/rstanarm/commit/7b88b5c3474f87af94b8de8349457be785087dd7#diff-611496f412cac947be720d17a0ee6d7463221d14731fbc18244756271e8f5189? If it's not needed anymore let's close this. I'm trying to clean up the lots of outdated PRs lying around in the rstanarm and rstan repos.

jgabry:
thanks!

PR #480, opened 2020-10-20
Fix backtransformation of intercept in stan_surv when prior_PD = TRUE

### LINKED ISSUE #479: The intercept is affected by the data when prior_PD = TRUE
#### Summary:

The intercept is back-transformed by `dot_product(xbar, beta)` even when `prior_PD = TRUE`.

#### Description:

When `prior_PD = TRUE` the covariates have no impact on the model fitting. Yet they have an impact on the intercept that is returned to the user, through their mean that is used in centering the intercept. Demonstrated by the reproducible example below.

Is this the desired behaviour? (my guess is probably not?)

Someone noticed this with `stan_surv`, but I think it might apply to every modelling function that back-transforms the intercept in some way.

#### Reproducible Steps:

```
library(rstanarm)

mtcars$wt1 = mtcars$wt * 10
fit1 <- stan_glm(
    mpg ~ wt1,            
    data = mtcars, 
    prior_intercept = normal(0, 10, autoscale = FALSE),
    prior = normal(5, 10, autoscale = FALSE),
    prior_aux = normal(0, 10, autoscale = FALSE),
    prior_PD = TRUE,
    refresh = 0,
)

mtcars$wt2 = mtcars$wt * -10
fit2 <- update(fit1, formula. = mpg ~ wt2)

fit1$coefficients['(Intercept)']
fit2$coefficients['(Intercept)']
```

#### RStanARM Version:

2.12.2 `feature/survival` branch. I haven't tried it on the master branch but the same [source code](https://github.com/stan-dev/rstanarm/blob/9cb331e634c5999132033eb4d97acf8e729632d8/src/stan_files/continuous.stan#L281) is there.

#### R Version:

4.0.2

#### Operating System:

Windows 10.

sambrilleman:
Intended to fix #479 for stan_surv models at least.

### DIFF
--- a/src/stan_files/surv.stan
+++ b/src/stan_files/surv.stan
@@ -1189,8 +1189,10 @@ generated quantities {
   // transformed intercept
   real alpha;
   if (has_intercept == 1) {
-    alpha = log_crude_event_rate - dot_product(x_bar, beta) + gamma[1];
+    if (prior_PD) alpha = gamma[1];
+    else alpha = log_crude_event_rate - dot_product(x_bar, beta) + gamma[1];
   } else {
-    alpha = log_crude_event_rate - dot_product(x_bar, beta);
+    if (prior_PD) alpha = 0;
+    else alpha = log_crude_event_rate - dot_product(x_bar, beta);
   }
 }


jgabry:
Looks good to me!

sambrilleman:
I'll hold off on merging this until we finish the discussion in the open issue thread (#479 and/or #480, which are each slight variations on a similar problem). I would want to keep `stan_surv` consistent with whatever happens in the other **rstanarm** modelling functions I guess.

sambrilleman:
I'm going to close this unmerged, taking @bgoodri's comments in https://github.com/stan-dev/rstanarm/issues/479 as the reason to not merge:

> The recentering always happens because the draw from the prior distribution of the coefficients is never exactly zero, although it is never far from zero under the default prior. I would have preferred no conditioning on the outcome variable at all, but Andrew forced us to for the book. I also agree that `stan_surv` is weirder because the prior predictive distribution it draws from is even less prior than for the other `stan_` functions, but I think the way it deals with the coefficients is fine.

But if anyone feels strongly about it going the other way, we could reopen and merge.

PR #492, opened 2020-12-10
Allow 0-column newdata if positive number of rows

### LINKED ISSUE #491: Passing empty tibble to posterior_predict() when using intercept-only models
#### Summary:
posterior_predict() (and related functions like posterior_epred()) should accept an empty tibble as the newdata argument 
when working with a model whose only right-hand side variable is the intercept.

#### Reproducible Steps:
``` r
suppressPackageStartupMessages(library(rstanarm))
suppressPackageStartupMessages(library(tidyverse))

fit_1 <- stan_glm(formula = mpg ~ 1,
                  data = mtcars,
                  refresh = 0)

# This works even though we don't "need" newobs to contain a value for cyl.

newobs <- tibble(cyl = 5)

pp <- posterior_predict(fit_1, newdata = newobs)

# This fails even though it should produce the same result as the first example.

newobs <- tibble()

pp <- posterior_predict(fit_1, newdata = newobs)
#> Error in .subset(x, j): invalid subscript type 'list'
```

<sup>Created on 2020-12-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jgabry:
fixes #491

This PR avoids performing one of the checks on `newdata` that resulted in an error if `newdata` had 0 columns. Allowing 0-column `newdata` is useful for doing `posterior_predict()` for intercept only models, as pointed out in #491  by @davidkane9. With this change, the following now works: 

```r
fit <- stan_glm(mpg ~ 1, data = mtcars)
posterior_predict(fit, newdata = tibble::tibble(.rows = 1)) # generate predictions for 1 obs
posterior_predict(fit, newdata = tibble::tibble(.rows = 2)) # generate predictions for 2 obs
...
```

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -814,7 +814,9 @@ validate_newdata <- function(object, newdata = NULL, m = NULL) {
     stop("NAs are not allowed in 'newdata'.", call. = FALSE)
   }
   
-  newdata <- drop_redundant_dims(newdata)
+  if (ncol(newdata) > 0) {
+    newdata <- drop_redundant_dims(newdata)
+  }
   return(newdata)
 }
 


jgabry:
This also works with `posterior_epred()` and `posterior_linpred()`. I will also add a test.

bgoodri (approved):

PR #493, opened 2020-12-11
matrix instead of special ppd class

### LINKED ISSUE #490: posterior_predict() and posterior_epred() return objects of different classes
#### Summary:
posterior_predict() and posterior_epred() return objects of different classes. Shouldn't they produce objects of the same class given that what they both return are draws from a posterior probability distribution?

#### Reproducible Steps:
``` r
suppressPackageStartupMessages(library(rstanarm))

fit_1 <- stan_glm(formula = mpg ~ cyl,
                  data = mtcars,
                  refresh = 0)

pp <- posterior_predict(fit_1)
class(pp)
#> [1] "ppd"    "matrix" "array"

pe <- posterior_epred(fit_1)
class(pe)
#> [1] "matrix" "array"
```

<sup>Created on 2020-12-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I much prefer the behavior of posterior_epred(). Matrices are easy to deal with. Objects of class ppd are not. But is there any reason for one function to return a matrix and one a ppd? That behavior makes teaching hard!

jgabry:
closes #490 

Drop superfluous class "ppd" from objects returned by `posterior_predict()`. This shouldn't break any CRAN packages, but we'll find out in reverse dependency checks and can revert if necessary.


bgoodri:
Why are we doing this?

jgabry:
Not because something is broken. Just that posterior_epred/linpred return simple matrices but for some reason posterior predict adds an unnecessary class (and adds it before "matrix"). according to #490 it can make teaching harder and it does seem totally unnecessary and sometimes requires unclass()-ing. Maybe an alternative to removing it (if there's a reason to keep it) is to just switch the order of classes to put matrix before ppd.

bgoodri:
Do we have any .pred methods? Or did we at one point? If not, I guess we
can get rid of the S3 classing.

jgabry:
We have `predictive_error.ppd` and `predictive_interval.ppd`, but this PR doesn't get rid of those (to avoid erroring if people have stanreg objects saved from previous versions of rstanarm with that class). This PR drops "ppd" from the class going forward (so it's just "matrix") and adds `predictive_error.matrix` and `predictive_interval.matrix`. I think that should prevent anything from breaking for anyone.

davidkane9:
Any updates on this?

jgabry:
> Any updates on this?

Sorry we never give you an update! I didn't see your comment until now. 

I think we're waiting on the next release of rstan before we submit the next rstanarm but I'm guessing we're on track for this summer at some point (@bgoodri would have a better estimate). 

But we don't need to wait until then to merge this. @bgoodri Do you prefer dropping the `"ppd"` class the way I currently have it in this PR would you prefer keeping the `"ppd"` class but moving it so `"matrix"` is the first class? Either would accomplish the goal of never having to unclass it.

bgoodri:
I think it can just be a matrix

PR #499, opened 2020-12-24
fix offset error for posterior_traj

### LINKED ISSUE #497: posterior_traj example errors
@sambrilleman I'm working on #496  and running r cmd check I noticed that this part of the `posterior_traj()` example

https://github.com/stan-dev/rstanarm/blob/aab5e594ba529ba50fc6495ed1e436d3633ebbb3/R/posterior_traj.R#L248-L249

fails with this error: 

```
Error in eval(predvars, data, env) : object 'logBili' not found
```

Same error happens with the next part of the example: 

https://github.com/stan-dev/rstanarm/blob/aab5e594ba529ba50fc6495ed1e436d3633ebbb3/R/posterior_traj.R#L270-L271

I think this is because `logBili` also needs to be in `newdata`. @sambrilleman I'm happy to fix this but I'm not sure what values `logBili` should have for these `posterior_traj()` examples. Should it just be set to its mean from the data? Or multiple different values?

jgabry:
fixes #497 

@pamelanluna @sambrilleman I think this is fixed now (the test I added works and the examples that had errors now also work) but can one or both of you take a quick look at this and make sure I didn't screw anything up? 

This can definitely wait until after the holidays if you're busy!


R/posterior_traj.R
```diff
@@ -401,7 +401,16 @@ posterior_traj <- function(object, m = 1, newdata = NULL, newdataLong = NULL,
     }
   }
     
-  newOffset <- model.offset(model.frame(terms(glmod), newX))
+  if (isTRUE(as.logical(glmod$has_offset))) {
```
jgabry, line 404: It seems that `glmod$has_offset` can be `1/0` instead of `TRUE/FALSE` for whatever reason

tests/testthat/test_stan_jm.R
```diff
@@ -520,6 +520,17 @@ for (j in c(1:30)) {
...
     }) 
     
+    test_that("posterior_traj works with new data with and without offset", {
+      nd <- pbcLong[1:2, ]
+      expect_s3_class(posterior_traj(f1, newdataLong = nd, dynamic = FALSE), "predict.stanjm")
+      
+      pbcLong2 <- pbcLong
+      pbcLong2$off <- 1
+      o <- SW(f1_off <- update(f1, dataLong = pbcLong2, formulaLong. = logBili ~ offset(off) + year + (year | id)))
+      nd <- pbcLong2[1:2, ]
```
sambrilleman, line 530: Do we need to drop the response column from the new data, to check that this test is passing when the new data doesn't contain the response (I think that was the original issue)?

We could extract just the `id`, `year` and `offset` columns I think, something like:

```
nd <- pbcLong2[1:2, c("id", "year", "off")]
```

tests/testthat/test_stan_jm.R
```diff
@@ -520,6 +520,17 @@ for (j in c(1:30)) {
       }  
     }) 
     
+    test_that("posterior_traj works with new data with and without offset", {
+      nd <- pbcLong[1:2, ]
```
sambrilleman, line 524: Same here I think, we might not want the response in the new data, so that we can catch the original error, something like:

```
nd <- pbcLong2[1:2, c("id", "year")]
```

R/posterior_traj.R
sambrilleman: Sounds good to me.

sambrilleman:
I just had a couple of minor comments on the test. But other than those, this all looks good to me.

Thanks for resolving this! It's much appreciated.

jgabry:
Thanks @sambrilleman. Good call on dropping the response variable in the test. I just pushed a commit fixing that.

jgabry:
Actually after fixing the test I'm now still seeing the error about missing logBili, so will have to look into that further.

jgabry:
Coming back to this. @sambrilleman Do you by any chance have time to look at this at some point? I'm not sure why we're still getting the error after the changes I made. Not sure if I made a stupid mistake somewhere or if it's more substantial.

jgabry:
To be more specific, it's this test that fails

https://github.com/stan-dev/rstanarm/blob/d28fafe0c24a083615af8704b95743021e5d47c1/tests/testthat/test_stan_jm.R#L528-L531

so it's the just the one where the model has an offset. Without the offset it's working ok.

jgabry:
Ok so I do think I solved this problem but I seem to have uncovered another related one: 

This line that gets run a bunch of times (because there's a loop over lots of models in the `stan_jm` tests)

```r
 o <- SW(mod_off <- update(mod, dataLong = pbcLong2, formulaLong. = logBili ~ offset(off) + year + (year | id))) 
```

works fine for some of the models but for others I get an invalid stanfit object because of this dimension mismatch error:

```
Error in new_CppObject_xp(fields$.module, fields$.pointer, ...) : 
  Exception: mismatch in dimension declared and found in context; processing stage=data initialization; 

variable name=y1_offset_auc; position=0; dims declared=(629); dims found=(9435) 
(in '/data/data_assoc.stan' at line 127; included from 'model_jm' at line 39)
```

PR #510, opened 2021-04-24
Fix compilation under rstan 2.26

andrjohns:
This PR adds `Makevars` flags needed to build `rstanarm` against the new 2.26 version of `rstan`


jgabry:
Thanks @andrjohns! Will this now _only_ work with 2.26 or will it still also work fine with older versions?

andrjohns:
Looks like I missed a comment from @hsbadr on this! This flag should only be defined if the associated `rstan` version is `>= 2.26`. ([that comment here](https://github.com/stan-dev/rstantools/pull/85#discussion_r619921248)). So this flag should be conditionally included using `if (utils::packageVersion("rstan") >= 2.26)`. I'll have a look into fixing this up

PR #511, opened 2021-05-06
Found a typo on roaches data docs

storopoli:
Guys, there is a typo on the roaches' data docs.

replacing "eldery" for "elderly"


jgabry (approved):
Thanks Jose!

Issue #512, opened 2021-05-06
Development version fails to install on Ubuntu 20.04


tmalsburg:
#### Summary:
Installing the development version fails on Ubuntu 20.04 with R 3.6.3.
  
#### Description:
The error message at the end of the trace is:
```
make: *** [/usr/lib/R/etc/Makeconf:175: stan_files/mvmer.o] Error 1
rm stan_files/jm.cc stan_files/binomial.cc stan_files/lm.cc stan_files/continuous.cc stan_files/bernoulli.cc stan_files/polr.cc stan_files/count.cc stan_files/mvmer.cc
ERROR: compilation failed for package ‘rstanarm’
* removing ‘/home/malsburg/R/x86_64-pc-linux-gnu-library/3.6/rstanarm’
Warning message:
In i.p(...) :
  installation of package ‘/tmp/RtmpayiOLb/file270684ca80d23/rstanarm_2.21.2.tar.gz’ had non-zero exit status
```

I have no problems installing other R packages.

#### Reproducible Steps:
- Start new R shell.
- Then enter 
``` R
Sys.setenv(MAKEFLAGS = "-j4") # change 4 to however many cores you can/want to use to parallelize install 
Sys.setenv("R_REMOTES_NO_ERRORS_FROM_WARNINGS" = "true")
remotes::install_github("stan-dev/rstanarm", INSTALL_opts = "--no-multiarch", force = TRUE)
```

#### RStanARM Version:
Current development version: dee0a2d45bf42b2df791072041151b753edd6af9

#### R Version:
3.6.3

#### Operating System:
Ubuntu 20.04.2 LTS

andrjohns:
If you haven't already, can you complete the installation instructions for RStan first ([link to those here](https://github.com/stan-dev/rstan/wiki/RStan-Getting-Started)) and verify that you can install RStan without any issues?

If you've already completed those steps, can you post more of the error log? Theres not enough there to see what the issue was

tmalsburg:
Sorry should have mentioned: Rstan (and brms and ulam) are working fine.

tmalsburg:
Just to be save, I reinstalled RStan following the instructions and verified that it's working (it is).

tmalsburg:
Error log:

[error.log](https://github.com/stan-dev/rstanarm/files/6435582/error.log)

Unfortunately not complete because my shell's buffer was too small.  Strangely I can't see any compiler errors, just warnings.

tmalsburg:
[...]

PR #513, opened 2021-05-07
Reduce recommended cores for makeflags during install

andrjohns:
As experienced over in #512, users can run out of RAM when installing `rstanarm` from source with the recommended `-j4`. This drops the recommended number to 2 and adds a warning


bgoodri:
Yeah

jgabry (approved):

PR #522, opened 2021-05-11
fix handling of empty interaction levels

### LINKED ISSUE #260: posterior_predict fails for binomial models with missing interactions
#### Summary:
Running posterior_predict on the output of an rstanarm::stan_glm estimation when there are interaction terms with missing levels causes an error. This only happened when my model was binomial.

#### Description:
I have a model I'm estimating where I want an interaction term between two categorical variables. Not all combinations of both variables are present, so when rstanarm estimates it lets me know that it dropped those levels (e.g. `In center_x(x, sparse) : Dropped empty interaction levels: group1B:group2c`). When I go to run posterior_predict I get the error: 
```
Error in stanmat[, beta_sel, drop = FALSE] : subscript out of bounds
``` 
because model.matrix creates a column with the empty interaction level, but the parameter matrix doesn't.

#### Reproducible Steps:
```
library(rstanarm)

group1 <- c("A", "B")
group2 <- c("a", "b", "c")
testFrame <- expand.grid(group1 = group1, group2 = group2)[1:5,]
testFrame$n <- 100
testFrame$theta <- runif(5)
testFrame$e <- round(testFrame$n*testFrame$theta)

fit <- rstanarm::stan_glm(cbind(e, n - e) ~ group1:group2, data = testFrame, 
                          family = binomial("logit"))
pred <- rstanarm::posterior_predict(fit)
```

[1] ‘2.17.2’

[1] ‘3.4.2’

#### Operating System:
SUSE Linux 11sp4

### LINKED ISSUE #425: Inconsistent handling of empty interations by stan_glm.fit and model.matrix.stanreg
#### Summary:
`stan_glm.fit` drops columns with empty interactions from the data, while `model.matrix` keeps them. Which leads to  `Error in stanmat[, beta_sel, drop = FALSE] : subscript out of bounds`
errors when running `posterior_predict` or `loo`.
#### Description:
`stan_glm.fit` runs the function `center_x`, which drops columns with empty interactions from the data. This is inconsistent with behavior of `model.matrix.default`, which keeps them as a column of zeroes and `glm` which keeps the empty interaction column and assigns a coefficient NA.

This leads to  `Error in stanmat[, beta_sel, drop = FALSE] : subscript out of bounds` when running `posterior_predict` or `loo` and I assume any other function that runs `get_x.default` since that is the function that calls the `model.matrix.default`.

Related to #260 

#### Reproducible Steps:
```{r}
d_empty <- data.frame(y = factor(c(1, 1, 1, 2, 2, 2)),
                      x = factor(c(1, 1, 2, 2, 3, 3)),
                      z = factor(c(1, 2, 1, 2, 1, 1)))

model.matrix(y~x*z, data=d_empty) # We see the column `x3:z2` consists of zeroes

fit <- stan_glm(y~x*z, data = d_empty, family = "binomial")
loo(fit)
# or
posterior_predict(fit)
```

#### RStanARM Version:
2.19.3

#### R Version and Operating System:
Both R version 3.6.2 with MacOS 10.15.3 and 
R version 3.6.0 with CentOS Linux 7 (Core)

#### Hacky fixes
I've managed to address the problem in two ways, but both solutions are hacky and I'm sure will lead to problems that I haven't foreseen.

The first is to add a function

```
drop_empty_levels <- function(x) {
  sel <- apply(x, 2L, function(x) !all(x == 1) && length(unique(x)) < 2)
  if (any(sel)) {
    # drop any column of x with < 2 unique values (empty interaction levels)
    # exception is column of 1s isn't dropped 
    warning("Dropped empty interaction levels: ",
            paste(colnames(x)[sel], collapse = ", "))
    x <- x[, !sel, drop = FALSE]
  }
  return(x)
}
```
and inside `model.matrix.stanreg` replace the line
`NextMethod("model.matrix")` with
`return(drop_empty_levels(model.matrix.default(object)))`.

This creates a different problem in that, if the data the data used to fit the model doesn't contain empty interaction levels, but we call `posterior_predict` with a `newdata` argument that's model.matrix will have empty interaction levels, then these will be dropped and we will still have an inconsistency in the number of columns.

The alternate method is to remove the lines from `center_x` that drop empty interaction levels.

```
xtemp <- xtemp[, !sel, drop = FALSE]
xbar <- xbar[!sel]
```
but this creates its own problem in that stan is fitting to a rank deficient matrix and will return a coefficient for the empty interaction level.

I think the correct solution probably involves dropping the empty interaction column for fitting, but having variable in the stanreg object and giving it an NA values for everything.

jgabry:
fixes #260
fixes #425

Fixes failures in `loo` and `posterior_predict` when there are empty interaction levels in the data. These levels were getting dropped for model fitting but not in `loo` and `posterior_predict`. 

To handle this we now carry around a `dropped_cols` slot in the stanreg object (for models that use `stan_glm.fit` or `stan_betareg.fit`) which can be accessed by `posterior_predict`, `log_lik`, etc., so they know which columns need to be dropped.

### DIFF
--- a/R/data_block.R
+++ b/R/data_block.R
@@ -23,16 +23,19 @@
 # @param warn Should warning be thrown if columns are dropped? 
 # @return A list with updated x and xbar.
 drop_empty_levels <- function(x, xbar = NULL, warn = TRUE) {
-  sel <- apply(x, 2L, function(w) !all(w == 1) && length(unique(w)) < 2)
+  sel <- apply(x, 2L, function(w) length(w) > 1 && !all(w == 1) && length(unique(w)) < 2)
   if (any(sel)) {
+    dropped_cols <- colnames(x)[sel]
     if (warn) {
-      warning("Dropped empty interaction levels: ", paste(colnames(x)[sel], collapse = ", "), 
+      warning("Dropped empty interaction levels: ", paste(dropped_cols, collapse = ", "), 
               call. = FALSE)
     }
     x <- x[, !sel, drop = FALSE]
     xbar <- xbar[!sel]
+  } else {
+    dropped_cols <- NULL
   }
-  nlist(x, xbar)
+  nlist(x, xbar, dropped_cols)
 }
 
 # Center a matrix x and return extra stuff
@@ -53,7 +56,10 @@ center_x <- function(x, sparse) {
   }
   
   dropped <- drop_empty_levels(xtemp, xbar)
-  nlist(xtemp = dropped$x, xbar = dropped$xbar, has_intercept)
+  nlist(xtemp = dropped$x, 
+        xbar = dropped$xbar, 
+        has_intercept, 
+        dropped_cols = dropped$dropped_cols)
 }
 
 # Deal with priors
--- a/R/log_lik.R
+++ b/R/log_lik.R
@@ -236,8 +236,11 @@ ll_args.stanreg <- function(object, newdata = NULL, offset = NULL, m = NULL,
   }
   if (is.stanmvreg(object) && !is.null(dots$stanmat)) {
     stanmat <- dots$stanmat # potentially use a stanmat with a single draw
-  }  
-  x <- drop_empty_levels(x, warn = FALSE)$x
+  }
+  
+  if (!is.null(object$dropped_cols)) {
+    x <- x[, !(colnames(x) %in% object$dropped_cols), drop = FALSE]
+  }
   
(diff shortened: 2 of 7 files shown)


jgabry:
Ok I think this is ready for review

djb17:
Is there a reason why this fix is still waiting to be merged?

jeliason:
Bumping. Is there any reason this PR cannot be merged?

bgoodri (approved):

PR #527, opened 2021-05-12
Updates stan_surv vignette to align with arXiv paper

sambrilleman:
This adds details to the `stan_surv` vignette, so that it aligns with the [arXiv paper](https://arxiv.org/pdf/2002.09633.pdf).


jgabry:
Thanks!

PR #550, opened 2021-09-24
Update Makevars for RStan >= 2.26

andrjohns:
This PR adds the Makevars changes needed for compatibility with the upcoming rstan release. The compiler/linker flags have been updated to match those provided by `rstantools`, and a conditional flag added (`-DUSE_STANC3`) which is needed for use with `rstan` >= 2.26

I've tested the build locally, but let me know if I've missed anything


andrjohns:
Obseleted by #574

PR #552, opened 2021-10-16
wells dataset wrong variable name

storopoli:
While lecturing this week on binomial regression I was showcasing the `wells` dataset (Gelman & Hill, 2007) and found an incosistency.
In the help pane the variable was named as `association` but it is only `assoc`.

```r
r$> names(wells)
[1] "switch"  "arsenic" "dist"    "assoc"   "educ"
```


storopoli:
@jgabry this is a simple typo fix. See if makes sense otherwise I can close this.

jgabry (approved):
Thanks @storopoli!

PR #514, opened 2021-05-07
csr_matrix_times_vector2 now uses csr_matrix_times_vector

SteveBronder:
Following the conversation on a [stanc3 pr](https://github.com/stan-dev/stanc3/pull/865#issuecomment-833650029) this modifies `csr_times_matrix_vector2` to have the same signature as the new stanc3 generated signature. It also just uses `csr_matrix_times_vector` since in the newest Stan math there is a reverse mode specialization for the `csr_matrix_times_vector` function.

PR #558, opened 2021-12-23
Remove custom C++ functions

hsbadr:
As discussed in #514 and #555, this PR removes all custom C++ functions and uses the replacement from Stan/Math. It also fixes building `rstanarm` with Stan v2.28.2, which requires a very minor patch for Math.


hsbadr:
I'll look into the test failure tomorrow; it seems to be related to the following exception:
```r
Exception: csr_matrix_times_vector: u/z (34) and v (35) must match in size
    (in '/model/eta_add_Zb.stan' at line 2; included from 'model_continuous' at line 150)
```

hsbadr:
@bgoodri Before I spend time to fix the failed examples/tests (which refer to the checks in `csr_matrix_times_vector`), are you ok with the change of removing the custom C++ functions to improve backward/forward compatibility with Stan/Math?

PR #597, opened 2023-09-09
Array syntax

bgoodri:
It is hard to know if I got 100% of the old arrays converted. The current canonicalizer in the Ocaml `stanc` will convert the top level .stan files, but it does not know what to do with the many files that only exist to be `#include`d into other Stan programs.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: rstanarm
 Type: Package
 Title: Bayesian Applied Regression Modeling via Stan
-Version: 2.21.4
-Date: 2023-04-03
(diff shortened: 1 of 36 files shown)


jgabry:
Nice. Any idea why tcrossprod is a problem? 

https://github.com/stan-dev/rstanarm/actions/runs/6128909508/job/16636479378?pr=597#step:8:421

bgoodri:
No, but it happens inconsistently on CRAN with the new rstan but without any change to the array syntax.

jgabry (approved):
Failures don't seem related to array syntax

PR #598, opened 2023-09-14
Update more deprecated syntax

andrjohns:
There are still some pieces of deprecated syntax in the rstanarm Stan models which will cause failures under 2.33:

- Forward declarations of Stan functions
- `fabs` deprecated in favour of `abs`
- `_cdf()` functions without `|` notation
  - Bug in 2.26 parser does not accept `cdf` functions *with* the `|` notation, so I've changed these to use the respective `_lcdf` function for compatibility with both


jgabry (approved):
Thanks @andrjohns!

PR #607, opened 2023-10-31
Bring feature/survival up to date with master

sambrilleman:
Brings the `feature/survival` branch in line with master. Also solves any errors, warnings, and solvable notes in the R CMD CHECK.


jgabry:
Thanks @sambrilleman!

sambrilleman:
I had dropped the CI build to only windows to save having to run all the builds. But turns out there is some of condition that skips the testthats on windows I think! Ha. So I think some tests might be failing but only surfaced on Linux/Mac. Hopefully I can find some time to look closer. But at least the r CMD check is passing all the other checks, so that's good at least. And at least it's in sync with master now too.

PR #616, opened 2024-02-17
Specify explicitly only the needed symbols to export from the DLL.

kalibera:
Specify explicitly only the needed symbols to export from the DLL.  This is safer wrt to possible naming conflicts and fixes linking on Windows using gcc 13.2, which produces weak symbols, which in turn cannot be exported from a DLL.

Also fixed R_init_rstanarm to be found by R.


jgabry:
Thanks for the PR. Gonna tag @bgoodri to take a look since he knows a lot more about this topic than I do.

bgoodri:
This is going to be fine. We need to test whether it continues to work on older Windows releases of R and if not, bump the minimum R version up to whatever it works with. The larger question is whether we also want to do this in rstantools after the next R release in April?

bgoodri:
Yes, if we enabled LTO for many, many other packages via rstantools, then we would generate the package's .def file when the `configure.win` script is called (which in turn calls `rstantools::rstan_config`).

The PR seems to fail to load rstanarm with the message:
```
Error: package or namespace load failed for ‘rstanarm’ in .doLoadActions(where, attach):
 error in load action .__A__.1 for package rstanarm: (function (ns) : object 'm' not found
```
This error comes comes from
https://github.com/stan-dev/rstanarm/blob/master/R/zzz.R#L20
Do we not need to call `Rcpp::loadModule` anymore?

kalibera:
Hmm, sorry, lets separate the two things. I've updated the pull request to just fix linking on Windows to export only selected symbols. I've removed my attempt to fix the registration from the patch, which is unrelated and will need further investigation.

kalibera:
> The PR seems to fail to load rstanarm with the message:
> 
> ```
> Error: package or namespace load failed for ‘rstanarm’ in .doLoadActions(where, attach):
>  error in load action .__A__.1 for package rstanarm: (function (ns) : object 'm' not found
> ```
> 
> This error comes comes from https://github.com/stan-dev/rstanarm/blob/master/R/zzz.R#L20 Do we not need to call `Rcpp::loadModule` anymore?

I don't know. The problem is non-deterministic on my system, sometimes it happens and sometimes not.

PR #632, opened 2024-09-26
Update Andrews todo

HPCurtis:
Remove for loop around the abs function now the abs function has been vectorised since 2.30


HPCurtis:
Very interested in how this failed as the model compiles fine and the issue is arising on the install of the rstanarm package.

jgabry (approved):
Thanks for the PR! I don't think this is failing due to anything you did, sometimes we've had problems with rstanarm installation on GitHub actions for reasons that, to be honest, I don't always understand. Your changes to the code look good!

bgoodri (approved):

PR #636, opened 2025-04-19
Fix mistaken ')' alignment in output test

MichaelChirico:
I'm not sure exactly how this test is passing now, but AFAICT the intent is to match `{n} (posterior sample size)` where `n` is the result of the `posterior_sample_size()` call.

Something like, `...` is swallowing the `"(...)"` string and the test passes because `500` is matched.


jgabry (approved):
Thanks for finding and fixing this. Indeed, I'm not sure how this was passing before!

Issue #637, opened 2025-05-21
findbars and friends are moving to the reformulas package


bbolker:
`findbars` and other formula-processing machinery has been moved to the `reformulas` package. In the current development version of `lme4`, these functions are no longer exported from `lme4`.  I plan to [partially revert this change](https://github.com/lme4/lme4/issues/830) so that `lme4::findbars` issues a deprecation warning ("please import this function from `reformulas` instead") rather than throwing an error, but in any case it would be good to move `rstanarm` to import these functions from `reformulas` rather than `lme4`.

I can submit a pull request if you'd like.

PR #639, opened 2025-05-22
migrate formula machinery from lme4 to reformulas

bbolker:
Responds to #637.

This does *not* pass tests cleanly but I don't think it's my fault ...


jgabry:
Thanks for the PR. I don't think the test failures are your fault either. The ones I see when running the tests seem unrelated to this change. This looks good to me and I'm going to approve this PR. Even though it's a very simple change to the code I'll wait to merge it so @bgoodri has a chance to take a quick look and at least find out that we're making this change and adding the new dependency.

jgabry (approved):

bgoodri (approved):

jgabry:
I also doubt it. For one, we have to use `@importFrom lme4` in many places do to import various generics from lme4 and then export our methods. For example, we have a lot of cases like this: 

https://github.com/stan-dev/rstanarm/blob/1250d1e5b2313c585d93e6a6ff8969a1c284d1f4/R/stanreg-methods.R#L206C1-L206C26

In addition to that `stan_(g)lmer` calls various lme4 functions (that are still in lme4 and not reformulas) like `glFormula` and `make_glmerControl`. I guess we could check for lme4 before using those functions, but since `stan_(g)lmer` is one of the most important functions in the package (perhaps the most important) it would be a bit strange not to make sure lme4 is installed when rstanarm is installed. 

From CRAN's perspective we could technically do this with `stan_(g)lmer, but I don't think we could do the stuff with the methods that I mentioned above (although maybe I'm wrong about that?).

jgabry:
Will go ahead and merge this since we both approved it.

PR #642, opened 2025-10-29
New pkgdown theme

VisruthSK:
Updating the theme to better match the main Stan page. This PR uses the new theme & use a GitHub Action to automatically build the dev version of the site on PRs and the release version on releases.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -22,3 +22,4 @@ dev-notes/*
 revdep/*
 
 ^\.github$
+^_pkgdown\.yml$


VisruthSK:
In _pkgdown.yml, 2 topics missing from index: "logit" and "pp_validate".
  Either add to the reference index, or use `@keywords internal` to drop from
  the index.
  
  Where should these two reference pages go? They're both exported so I presume the documentation should be published. @jgabry

VisruthSK:
I think this PR only needs a redocument and it should be good. I'm trying to do so now.

VisruthSK:
Math seems to be fixed at https://mc-stan.org/rstanarm/articles/lm.html, [rerunning](https://github.com/stan-dev/rstanarm/actions/runs/20108187082) the build after your changes. Was there anything else that needed to be fixed?

jgabry:
I added `_pkgdown.yml` to `.Rbuildignore` to avoid an R cmd check NOTE but otherwise I think this is ready to go.

PR #653, opened 2026-07-24
Move shinystan to suggests and bump Roxygen

### LINKED ISSUE #623: Request: move `shinystan` from `Imports` to `Suggests`?
#### Summary:
`rstanarm` imports `shinystan`, meaning that all packages that either suggest or import `rstanarm` also have to install a whole lot of `shiny`-related packages. This can cause issues in CI and for other package maintainers.

#### Description:
Hello, I don't use `rstanarm` but I maintain a package that `Suggests` it ([`datawizard`](https://cran.r-project.org/web/packages/datawizard/)). Recently, there were some failures in CI due to installation errors for [`QuickJSR`](https://cran.r-project.org/web/packages/QuickJSR/). I was a bit intrigued because our package has nothing to do with Javascript and doesn't use any `shiny`-related packages. It turned out that `QuickJSR` was installed because `shinystan` is also installed, and that's because we suggest `rstanarm`:
``` r
pak::pkg_deps_explain("datawizard", "QuickJSR", dependencies = TRUE)
#> ℹ Loading metadata database✔ Loading metadata database ... done
#> datawizard -> brms -> rstan -> QuickJSR
#> datawizard -> rstanarm -> rstan -> QuickJSR
#> datawizard -> rstanarm -> shinystan -> rstan -> QuickJSR
```
(note that `QuickJSR` is also installed by `rstan` but this encouraged me to open this issue)

As I said before, I never used `rstanarm` before but it seems to me that `shinystan` isn't so important here. From what I can see `shinystan` is only imported to provide a method for objects of class `stanreg`. However, this doesn't require importing `shinystan`, suggesting it would be enough using [`s3_register()`](https://vctrs.r-lib.org/reference/s3_register.html) (see in particular the section "Usage in other packages").

I didn't find a trivial way to compute how many mandatory dependencies would be removed if `shinystan` was moved from `Imports` to `Suggests`, but I'd assume this number is quite high given all the dependencies of `shiny` that would no longer need to be installed.

``` r
library(deepdep)

dd <- deepdep("rstanarm", depth = 2)
plot_dependencies(dd)
```

Is this something you would consider?

#### Reproducible Steps:
N/A

#### RStanARM Version:
2.32.1

#### R Version:
4.4.0

#### Operating System:
Windows 10 x64 (build 19045)

VisruthSK:
This PR was made with assistance from ChatGPT.

Move shinystan to suggests and defining a small compat helper to ensure the API doesn't change. Also bumps roxygen to v8, deleted test plots artifact, and forced lf normalization.

Fixes #623.


R/launch_shinystan.R
```diff
@@ -73,46 +73,61 @@
...
-#' 
+#'
 #' # Prevent ShinyStan from preparing graphical posterior predictive checks that
 #' # can be time consuming. example_model is small enough that it won't matter
 #' # much here but in general this can help speed up launch_shinystan
 #' sso <- shinystan::as.shinystan(example_model, ppd = FALSE)
 #' if (interactive()) launch_shinystan(sso)
 #' }
 #' }
+launch_shinystan <- function(object, ...) {
```
jgabry, line 108: This is already a generic in the shinystan package, why do we need to define it here? Can checking requireNamespace not go inside `launch_shinystan.stanreg`? Maybe I'm missing an obvious reason?

jgabry:
> Move shinystan to imports 

To suggests not imports. Thanks for doing this. See my one review comment, which is really a question

R/launch_shinystan.R
VisruthSK: We only need the generic here if we want rstanarm::launch_shinystan() to still work. If we're fine with breaking that we can delete it. Since shinystan is import now we can own the warning when people don't have shinystan installed or let R throw default namespace not found error.

The S3 method doesn't need the require namespace since it won't work unless shinystan is installed to register the generic, iirc.

jgabry: > Since shinystan is import now 

This PR moves it to suggests, but I see what you mean about the S3 method.

jgabry:
We're getting this error: 

> Namespace dependency missing from DESCRIPTION Imports/Depends entries: 'shinystan'

R/launch_shinystan.R
VisruthSK: Oops sorry no idea how I got that backwards twice. Fixed in main issue

jgabry:
> We're getting this error: 

>> Namespace dependency missing from DESCRIPTION Imports/Depends entries: 'shinystan'

https://github.com/stan-dev/rstanarm/actions/runs/30084028618/job/89451905454?pr=653#step:7:84

VisruthSK:
Fixed, and fixed some roxygen messages while I was at it. Just some linebreaks in import statements it didn't like.

VisruthSK:
I think these errors are unrelated: https://github.com/stan-dev/rstanarm/actions/runs/30122362006/job/89577779271#step:7:1

Though should be fixed of course.

PR #657, opened 2026-09-19
Fix current test errors, update to testthat 3rd edition

andrjohns:
Several tests were failing due to changes in the outputs from other packages that were being compared against (e.g., `lme4`) as well as the reliance on deprecated syntax from `ggplot2`.

I've also updated the actions workflow to use the built-in concurrency handling (instead of the custom `n1hility/cancel-previous-runs` workflow) and automatically detect and install all suggested deps instead of the hand-maintained list.

This PR also updates the test syntax for testthat 3e, enabling running the tests in parallel.

I've tried to keep the changes grouped by commit so it's easier to review, plus an (AI-generated) summary below, but let me know if it would be better split into multiple PRs!

## Summary

This branch gets CI running again and updates the package and its tests for newer versions of its dependencies (lme4, loo, bayesplot, ggplot2, testthat and roxygen2).

- **`VarCorr.stanreg`** (`R/stanreg-methods.R`): each matrix now has class `c("vcmat_us", "matrix", "array")`, to match what current lme4 returns.
- **`get_common_cnms`** (`R/jm_data_block.R`): the result is now sorted by name, so the grouping-factor order is the same in every submodel of `stan_jm`/`stan_mvmer`.
- **`pp_check`** (`R/pp_check.R`): `ppc_error_scatter_avg` now gets the default `x = 1:length(y)`, like `ppc_intervals`, `ppc_ribbon` and `ppc_error_binned` already did.
- **`plot_nonlinear`** (`R/stan_gamm4.R`):
  - `aes_()` (deprecated) is replaced by `aes()`, and `size` by `linewidth` for lines and ribbons.
  - Setting the variables to `NULL` locally removes the R CMD check "no visible binding" NOTEs.
- **testthat edition 3, run in parallel:**
  - Dropped `context()`.
  - `expect_equivalent` / `check.attributes = FALSE` became `ignore_attr = TRUE`.
  - `tol =` became `tolerance =`.
  - `expect_is` became `expect_s3_class`, `expect_type` or `inherits()`.
- **loo:** the expected error text is now "All models must have the same number of observations". The helper passes `r_eff = TRUE` to match the implementation.
- **`posterior_predict`:** the helper had a typo, `regexep`, so the error message was never checked. It is now `regexp` with the current message.
- **pp_check:** the newer bayesplot functions (`ppc_calibration*`, `ppc_dots`, `ppc_loo_calibration*`, `ppc_loo_pit_ecdf`) are excluded from the generic loop.
- **VarCorr:** attribute names are sorted before comparing.
- **clogit:** the reference values that depended on the StanHeaders version were removed, and `ITER` goes from 100 to 200.


andrjohns:
Unfortunately, the mac R-devel workflow will still fail until this PR is merged upstream: https://github.com/r-lib/actions/pull/1108

jgabry:
Thanks @andrjohns, this is great. I’ll take a closer look soon.

jgabry (approved):
Thanks again @andrjohns. I added you to the DESCRIPTION file (I realized we hadn't added you as a contributor yet) and I'll let CI finish before merging to see if Mac r-devel passes now that your r-lib PR was merged

